PR #2, opened 2021-06-15
Dckesler/docusaurus

dckesler:
This includes adding the docusaurus dependencies and build processes, and converting the getting-started, and celo-holder-guide docs to docusaurus.


critesjosh:
What would you like reviewed exactly? 

I am able to run the docs site locally with no problem. There are some bad links, like the "edit this page" link at the bottom of each page and unnecessary gitbook references, but these are simple fixes that can be addressed later.

dckesler:
Okay sure let me fix the linter and then if you think the docs look good with what I've done so far then we should be good to merge.

critesjosh (approved):
this is a great start. it works for me locally, so let's go from here

PR #12, opened 2021-06-17
Dckesler/page ref

dckesler:


src/plugins/aliases.ts
```diff
@@ -0,0 +1,16 @@
+const path = require('path');
```
dckesler, line 1: This plugin allows us to add in webpack aliases. In this case I'm adding `@components/` that is an alias for `src/components/` and can be referenced in our mdx files.

tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
dckesler, line 1: Stuff so we can write our components in typescript.

yarn.lock
```diff
@@ -1238,6 +1238,11 @@
     url-loader "^4.1.1"
     webpack "^5.28.0"
 
+"@docusaurus/module-type-aliases@^2.0.0-beta.0":
```
dckesler, line 1241: Typescript dev dependencies.

src/components/PageRef.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
dckesler, line 1: This is the actual page ref component. It's pretty simple.

sidebars.js
```diff
@@ -81,9 +81,5 @@ module.exports = {
         'celo-holder-guide/celo-recovery',
       ]
     },
-    {
```
dckesler, line 84: I just now noticed this wasn't a Celo doc it was a docusaurus thing.

docusaurus.config.js
```diff
@@ -85,4 +88,7 @@ module.exports = {
       },
     ],
   ],
+  plugins: [
```
dckesler, line 91: Here's where I add the aliases plugin.

package.json
```diff
@@ -32,5 +32,13 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
```
dckesler, line 36: This is typescript stuff

critesjosh (approved):
looks good!

PR #27, opened 2021-07-12
deprecated use of Gold

Nanunaki:
Changed mention of Locked Gold to Locked Celo in docs on Governance page.


joenyzio:
See [430](https://github.com/celo-org/docs/pull/430)

PR #28, opened 2021-08-06
Updating one doc changed since fork from monorepo

ericnakagawa:
Doc updates that account for changes since the docs were forked for docusaurus transition


critesjosh (approved):
lgtm!

PR #37, opened 2021-08-23
doc: Update new foundation policy

YazzyYaz:


critesjosh (approved):
updated the title block and the "info" notes to conform to the docusaurus standard. otherwise looks good.

PR #56, opened 2021-09-10
Add angolia supported search

critesjosh:
Add search to the site.


critesjosh:
I don't think the site has been indexed yet, so I don't think the search functionality will actually work yet. But the search bar is there and should be configured correctly.

ericnakagawa (approved):
LGTM

PR #60, opened 2021-09-13
Migrate keystores walkthrough docs

eelanagaraj:
Keystores walkthrough from a couple of months ago was missing, this should add it to the `developer-resources/walkthroughs`.

### Testing
- Built locally and checked that links work


critesjosh (approved):
thanks for adding!

PR #67, opened 2021-09-13
Update Using MetaMask with Celo

joenyzio:
Update Using MetaMask with Celo to match updates previously made to Docs 1.0


ericnakagawa (approved):
LGTM

PR #129, opened 2021-09-30
Add metamask buttons

critesjosh:
This PR adds the ability of docs pages to detect Metamask. It also adds buttons that allow readers to add the celo networks and tokens (cUSD and cEUR) to metamask from directly within the docs.


ayaxdilawar (approved):

joenyzio:
Closing but can use as example if we decide to add metamask buttons to docs in the future.

karlb:
Do we intend to merge this? It has been reopened without explanation or further work, so I am not sure.

PR #131, opened 2021-10-01
Update web-app developer guide

jsaur:
Hi, this is my first PR off the Celo docs :)
Essentially I went through the web-app flow described here (https://docs.celo.org/developer-guide/start/web-dapp) and ran into some issues. I've made 3 separate commits to capture the 3 issues I ran into.

1) When I attempted the "Connecting to the user's wallet" section, I got the error "Module not found: Can't resolve 'fs'". Searching the discord channel @AlexBHarley mentioned copying the next.config.js found here: https://github.com/celo-org/use-contractkit/blob/master/packages/example/next.config.js - I used that as the starting point but simplified it a bit for this demo to the minimum required to get things working.

2) I ran into a new error after fixing the next.config.js which was "Error: Components using the use-contractkit hook must be a child of a ContractKitProvider". For that I looked at the example here: https://github.com/celo-org/use-contractkit#wrap-your-application-with-contractkitprovider - and copied the pattern of adding a WrappedApp. Note that when there was not default export I saw the error "Error: The default export is not a React Component in page" so I export the WrappedApp on the last line.

3) I ran into a simple error around a missing dependency "useState".

After these updates everything else ran smoothly (I didn't try the optional Locking Celo section yet).

Let me know if you have any other suggestions, or if there are simpler steps to get everything working that I missed.


critesjosh (approved):
looks good! thanks for the contribution!

PR #132, opened 2021-10-04
Add node version

jsaur:
Hi, I thought it would be useful to include the Node version here. I see it now in other parts of the docs, but this was one of the first ones I tried and was using Node 14. The celo ganache-cli fails with "Error: callback already called", which according to https://support.chainstack.com/hc/en-us/articles/900001783126-Ganache-CLI-fails-with-Error-callback-already-called- is due to running an older version of ganache against Node 14. At some point it would probably be nice to update https://github.com/celo-org/ganache-cli to be in line with the latest ganache, but in the short run specifying Node 12 works fine.


critesjosh (approved):
thanks for clarifying.

PR #136, opened 2021-10-07
Bridging docs

joenyzio:
Added bridging native assets, bridging to celo, and bridging tokens with etherscon to Developers > Bridging section.


ericnakagawa (approved):
Thank you Joe!

PR #218, opened 2021-11-03
Update number on the ultralight sync protocol

nategraf:
For increased clarity, this PR updates the parameters mentioned on the ultralight sync page to match mainnet


critesjosh (approved):
👍

PR #220, opened 2021-11-05
Fix links

joenyzio:
Updated 404 links identified by SEO Moz and updated others to reference links.


joenyzio:
Build is broken on Spanish page after fixing link that was previously broken on English page.

critesjosh (approved):
lgtm

PR #230, opened 2021-11-11
cask updates

jsaur:
The command `brew cask install <package>` was failing for me. It looks like the syntax has been updated to `brew install --cask <package>`: https://stackoverflow.com/questions/30413621/homebrew-cask-option-not-recognized


critesjosh:
based on the first comment in the stack overflow answer, it looks like we could just remove references to `cask` altogether. i just tried a couple without `--cask` and it worked

jsaur:
Sounds good, ya it works for me too without --cask

critesjosh (approved):
👍

PR #237, opened 2021-11-16
Brand update

joenyzio:
Celo.org updated to darker look. Docs dark mode updated to match new look.


joenyzio:
Build failing on ValidationError: "colorMode.backgroundColor" is not allowed

critesjosh (approved):
looking good!

PR #245, opened 2021-11-30
Valora | Account Key = Recovery Phrase

joenyzio:
Updated all improperly named instances of Account Key to Recovery Phrase in docs that refer to Valora.


critesjosh (approved):
lgtm

PR #249, opened 2021-12-03
Adds a Tutorial for Migrating to v1

ericnakagawa:
Two parts: 1) adds new tutorial and stores images in static folder
2) does some very light clean up and legibility fixes to other bridging docs. primarily adds a code block around ethereum addresses, but this should be expanded to the entire aite.


critesjosh (approved):
lgtm!

PR #255, opened 2021-12-06
Create 2021-12-06-addtokentocelowallet.md

vissequ:
Hello friends


critesjosh:
thanks for this @vissequ! It looks like you added images using HTML. Can you update them to use markdown syntax? it is like a link in markdown with a `!` before it, so `alt text`

critesjosh (approved):
👍 thanks for the contribution!

PR #275, opened 2021-12-17
Update bridging-tokens-with-etherscan to optics V2 addresses

critesjosh:


bcamacho (approved):
Validated addresses using merge PR 732 -> https://github.com/celo-org/optics-monorepo/tree/main/rust/config/production-community/bridge/1637963999617

PR #277, opened 2021-12-22
docs: added await for promise

cajubelt:
In the example code, there should be an `await` before `accountWrapper.getDataEncryptionKey` since it returns a promise


critesjosh (approved):
👍

PR #286, opened 2022-01-10
Top level nav

joenyzio:
Reorganized some top level nav.

- Broke apart the Welcome section into Basics and Use Celo. Welcome was turning into a catch-all for topics instead of an intro.
- Linked blog to Celo Medium and turned original blog link into Tutorials section. Thinking we could pull tutorials from other sections into here to clean things up a bit. 
- Brought Integration into Develop section. Wasn't much material here.


critesjosh (approved):
👍

PR #290, opened 2022-01-13
Update erc20-transaction-fees.md

nategraf:
Remove "Note that this number is expected to drop significantly in the near future." as the gas cost was already reduced and we have no immediate plans to reduce it further.

Also, capitalize the section heading


critesjosh (approved):
👍

PR #296, opened 2022-01-28
Added RedStone oracles docs

hatskier:


critesjosh:
thanks for the PR. I wasn't aware of Redstone, but it looks really interesting. Do you have standard reference contracts deployed to each EVM chain, or do dapps just need to use the `PriceAware` library?

Also curious about what dapps on other EVM chains are using your data feeds

hatskier:
> thanks for the PR. I wasn't aware of Redstone, but it looks really interesting. Do you have standard reference contracts deployed to each EVM chain, or do dapps just need to use the `PriceAware` library?
> 
They just need to extend the `PriceAware` contract for now. Actually, it will be converted to library soon :)

> Also curious about what dapps on other EVM chains are using your data feeds
We are just finalising our contracts audits and our initial infrastructure configuration. We are working on integration with:
- Moola (lending protocol on Celo)
- Cybertime (NFT marketplace on Celo)
- [Deltaprime](https://deltaprime.io/)
- [TracerDAO](https://tracer.finance/)

We also have few implemented a and tested POC apps like for example https://github.com/redstone-finance/komodo-celo

We are providing data via HTTP API for about a year now and we have ~400000 GET requests daily. But we are just launching with providing data on EVM compatible blockchains.

Thinking now, maybe it makes sense to add a note to the docs, that before using in production apps with high volume, we would like to finalise the audits and help with the correct integration :)

critesjosh:
very cool!

> Thinking now, maybe it makes sense to add a note to the docs, that before using in production apps with high volume, we would like to finalise the audits and help with the correct integration :)

Do you want to add this and then I will approve?

hatskier:
> very cool!
> 
> > Thinking now, maybe it makes sense to add a note to the docs, that before using in production apps with high volume, we would like to finalise the audits and help with the correct integration :)
> 
> Do you want to add this and then I will approve?

Just added :)

critesjosh (approved):
lgtm! 👍

PR #302, opened 2022-03-02
Update celo-stack.md

rony23-byte:


critesjosh (approved):
thanks for the clarification!

PR #305, opened 2022-03-07
Link 10DLC guidance

eelanagaraj:
Updates our 10DLC guidance and docs. For now, it links the existing Notion pages so that new validators can find this information. Also removes the configuration of country defaults in favor of just linking to the config (I am still concerned that de-duplicating this will unfortunately mean that people won't find it as easily, but at least the information will only need to be kept up-to-date in one place).

# Related
Closes celo-org/celo-monorepo#9052


0xarthurxyz (approved):
Brilliant thanks! 🙏

PR #308, opened 2022-03-30
Updated Contractkit v2 migration with info about attestations wrapper 

aaronmgdr:
I missed that I had made a breaking change to the getConfig function of Attestations


critesjosh (approved):
lgtm

PR #312, opened 2022-04-14
Update addToken function behavor In the docs

ninabarbakadze:


critesjosh:
just curious, is this part just no longer required for adding an asset?

m-chrzan:
> just curious, is this part just no longer required for adding an asset?

Exactly. This was removed in https://github.com/celo-org/celo-monorepo/pull/9367 to make new stable token deployments smoother.

critesjosh (approved):
👍

PR #336, opened 2022-06-21
Documentation on how to run an oracle

martinvol:


martinvol:
I had too many build issues, so will start from scratch

PR #340, opened 2022-06-27
Update link to matchmaking example

jessan-clabs:
Small fix to docs.


nestorbonilla:
Could you update the link, please? It seems it has been updated, and the proposed one is not working.

joenyzio:
Proposed link is broken.

PR #345, opened 2022-07-01
Add Celo Tracker to Celo Dapp list

gnardini:
Added Celo Tracker's info to Celo's Dapp gallery


nestorbonilla:
PR solved at https://github.com/celo-org/docs/pull/426

PR #402, opened 2022-07-26
Added dapps to showcase

0xBex:
Added Santym, Savings Circle, Bienvenir, Cryptum Woocommerce Checkout, and Muggle Pay to showcase page from Gallery


nestorbonilla:
Added on https://github.com/celo-org/docs/pull/421

PR #409, opened 2022-08-17
Update voting-validators.md

vissequ:
Changing where it says there is a maximum of 100 Validators to "110 Validators"


joenyzio:
https://github.com/celo-org/docs/pull/420

PR #427, opened 2022-09-06
update validator specs, add a section for proxy or full node

ericnakagawa:
I've added two sections one for validator nodes, and one for proxy/full nodes.

The intention here is to inform potential node operators to properly size their estimates.


diwu1989:
For the CPU, I recommend we give a bit more guidance on the type of CPU that is necessary. For example, there are older CPU from prior generation that fit the 3GHZ number but actual instruction per clock is much lower than modern CPU.

Suggestion is:
```
4 core / 8 thread CPU with 3ghz on modern CPU architecture newer than 2018
Intel Cascade Lake or Ryzen 3000 series or newer
Geekbench 5 Single Threaded score of >1000 and Multi Threaded score of > 4000
```

diwu1989:
I would also strongly recommend we ask every validator (or maybe even as a requirement for Celo cohort foundation votes) that they run the validator in a carbon neutral or low-carbon data center.

For example, Google Cloud has a few designated low-carbon locations
image

Likewise, Hetzner has 100% renewable energy in Finland.

nestorbonilla (approved):

PR #467, opened 2022-10-26
Path for Celo Randomness doc changed

viral-sangani:
@joenyzio  Eric asked to change the path to randomness doc from `Protocol > Identity > Celo Randomness` to `Protocol > Celo Randomness.`


ericnakagawa (approved):
LGTM

PR #465, opened 2022-10-26
Celo Sage Hompage

joenyzio:
Moved notion to docs

PR #468, opened 2022-10-26
Celo Announcement article migrated from Hashnode to Celo docs

viral-sangani:
@joenyzio 
- Migrated the article from Hash node
- Added `Celo Sage` tag
- Don't delete this branch yet; I will add the Hero image once Lyle has it ready for Celo sage.


joenyzio:
Included in https://github.com/celo-org/docs/pull/465

PR #486, opened 2022-11-17
Plumo updates

joenyzio:
- Removed Ultralight Sync page
- Replaced with Consensus > with Plumo Ultralight Sync
- Removed MPC Contributor instructions
- Added link to Plumo Blog
- Replaced Block number example in blog with Balance verification example


sidebars.js
```diff
@@ -749,10 +749,11 @@ const sidebars = {
               label: "Locating Nodes",
               id: "protocol/consensus/locating-nodes",
             },
+            //remove ultralight sync from docs
             {
               type: "doc",
-              label: "Ultralight Sync",
-              id: "protocol/consensus/ultralight-sync",
```
ericnakagawa, line 755: please add a redirect from https://docs.celo.org/protocol/consensus/ultralight-sync to the new page

PR #511, opened 2022-12-09
Update Hosted Nodes documentation

jcortejoso:
Updates for the `Hosted Nodes` documentation.
Changes:

- Delete AWS documentation until images are regenerated using latest base images and fixed scripts (and we validate everything works)
- Updated gcp images
- Added reference to disk and instance customization


lvpeschke (approved):
Thank you 🚀

PR #579, opened 2023-01-19
Tutorial page updates

viral-sangani:
New feat added - 
- Back button on all tutorial pages
- Remove side bar that shows random article list
- Added table on content


nestorbonilla (approved):
LGTM

PR #610, opened 2023-01-31
Fix Links

aaronmgdr:
Updates all Faucet Links to go to Faucet.celo.org

also updates celo.org/build  links to celo.org/developers 

build was essentially the developers page a long time ago and no longer exists


nestorbonilla (approved):
LGTM

PR #642, opened 2023-02-08
0xarthurxyz/remove attestation service requirements

0xarthurxyz:
Removes attestation service requirements for mainnet and baklava validators in `Validators > Running a Validator`.
Doesn't remove identity sections in other parts of documentation (quick fix to avoid validator confusion).


nestorbonilla (approved):
LGTM

PR #660, opened 2023-02-15
suggest using latest version of composer

aaronmgdr:
by default. npx will use a cached version of the package if one exists. This means if someone tried composer long ago they will use that old version and likely will have a bad time. :(


nestorbonilla (approved):
LGTM

PR #735, opened 2023-03-06
Include CIP-29 in uptime score calculation

karlb:
[CIP-29](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0029.md) introduces a grace period which should be mentioned here.


karlb:
I only put this into draft mode because I'm not sure if CIP-29 is active on mainnet, yet. If you know that it is, feel free to mark this as ready for review!

palango (approved):

PR #769, opened 2023-03-16
fix error

ericnakagawa:
bug in docs


ericnakagawa:
Also, this code block doesn't work:

```
await deploy(“CeloMarketPlace”, {
from: deployer,
args: [],
log: true,
})

module.exports.tags = ["CeloMarketPlace];
```

Needs to be: 

```
module.exports = async ({ getNamedAccounts, deployments }) => {
    const { deploy } = deployments;
    const { deployer } = await getNamedAccounts();

    await deploy("CeloMarketPlace", {
        from: deployer,
        args: [],
        log: true,
 
    });

};
module.exports.tags = ["CeloMarketPlace"];
```

developerkunal (approved):
Approve @Ishanpathak1

PR #814, opened 2023-03-21
Fix Slug

developerkunal:
Fixing Slug


Ishanpathak1 (approved):
Approved @developerkunal

PR #913, opened 2023-04-11
Update mainnet node requirements from 256GB to 4TB

bcamacho:
Currently, the mainnet is nearly two terabytes of data in size. Modifying the recommended size to 4TB for nodes to sustain storage over time.


therealharpaljadeja:
These requirements for Archive nodes, so closing.

PR #1060, opened 2023-05-12
Update celo-foundation-voting-policy.md

ericnakagawa:


therealharpaljadeja:
Please re-open if important

PR #1093, opened 2023-07-07
Update gitcoin bounties link

karlb:
The old link 404s and this looks like the right replacement.


karlb:
Great, the new link is dead now, too. Do we want to keep using Gitcoin bounties at all?

PR #1099, opened 2023-07-20
Add explanation of gas price for alternative fee currencies.

aaronmgdr:
see https://github.com/wagmi-dev/viem/discussions/914

### DIFF
--- a/docs/developer/viem/index.md
+++ b/docs/developer/viem/index.md
@@ -37,4 +37,21 @@ The TLDR is that passing a celo chain from `viem/chains` into the config of `cre
 
 ```
 
+## Gas Price
+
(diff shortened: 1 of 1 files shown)


docs/developer/viem/index.md
```diff
@@ -37,4 +37,21 @@ The TLDR is that passing a celo chain from `viem/chains` into the config of `cre
 
 ```
 
+## Gas Price
+
+When paying for transaction with an alternate feeCurrency token it is important to know the price of gas denominated in that token. As such Celo nodes accept an optional param of the address of the token for the `eth_gasPrice` call. Therefore rather than use viem's `publicClient.getGasPrice()` function you should fetch it like the example.
```
karlb, line 42: Is it clear enough that you don't have to do this if you are fine with the default gas price handling? It sounds a bit like you have to do it every time you specify a feeCurrency.

karlb (approved):

docs/developer/viem/index.md
karlb: Ah, I assumed they had custom gas price fetching. Makes sense then!

PR #1101, opened 2023-07-29
removing multichain

Tjaaark:
CEO in jail, bridge hacked, front end still live - better to remove the link


jcortejoso (approved):

jcortejoso:
thanks for the pr

PR #1131, opened 2023-10-02
Release process update

pahor167:
Added verification step into release process.


viral-sangani:
@pahor167 Since Truffle is being sunset, can we add steps for verifying via Hardhat?

PR #1174, opened 2023-12-14
adds envio to celo showcase as supported indexer

SvenMuller95:
Envio <> Celo


therealharpaljadeja (approved):
Thank you for your contribution

PR #1177, opened 2023-12-22
Update client release docs

palango:
- Remove some specifics, such as release schedule or fixed distribution schedule.
- Remove mentions of developer signed artifacts
- Remove mentions of prebuilt binaries, as we currently don't supply those


therealharpaljadeja (approved):
LGTM!

PR #1276, opened 2024-04-04
fix: move `holder` to top level in the sidebar

therealharpaljadeja:


docusaurus.config.js
```diff
@@ -120,17 +120,18 @@ module.exports = {
...
-              // position: "left",
             },
             {
               to: "protocol/",
               label: "Protocol",
-              // position: "left",
+            },
+            {
+              to: "holder/",
+              label: "Holder",
```
lvpeschke, line 130: I think I would have expected the `Holder` section to be at the same level as `Developer` and `Validator`, wdyt?
Especially judging by the content on https://docs.celo.org/holder, which reads like a big entry page.

therealharpaljadeja: `Developer` and `Holder` are at the same level we just tucked `Holder` into the more part in order to have a better mobile responsible navbar.

If we want `Holder` to be outside the dropdown then we have to trade it with `Validator` or `What is Celo` item.

image

image

lvpeschke: That makes sense. I wouldn't trade it for one of the other options then.
How about making it the first in the dropdown?

therealharpaljadeja: Done!

lvpeschke (approved):

PR #1332, opened 2024-06-05
fix: broken celo etherscan api link

Another-DevX:
On the docs the celo-alfajores api link are the same than celo api, that's actually incorrect https://api.celoscan.io/api
This is the correct api endpoint according hardhat verify docs.


therealharpaljadeja (approved):
Thank you for your contribution

PR #1345, opened 2024-06-26
chore: update links.yml config

lvpeschke:
Running the workflow manually (see https://github.com/celo-org/docs/actions/runs/9679044164/job/26704145144) fails on creating a ticket because the body is too big (see https://github.com/peter-evans/create-issue-from-file/issues/1049).

This PR aims to reduce the amount of failures to be reported by
- adding caching and a GITHUB_TOKEN as per instructions at https://lychee.cli.rs/github_action_recipes/caching/#caching-in-github-actions for speed
- removing the progression indication (recommended for CI)
- adding verbose logging

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -25,7 +25,6 @@ jobs:
         uses: lycheeverse/lychee-action@v1
         with:
           args: >-
-            --base .
             --verbose
             --no-progress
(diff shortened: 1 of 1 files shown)


therealharpaljadeja (approved):
Thank you for the fix!

PR #1348, opened 2024-06-28
fix (docs): update `code-library.mdx` to correct `window.ethereum` ob…

khalifa47:
…ject

Object should have the boolean value `window.ethereum.isMiniPay` instead of `window.ethereum.isMinipay`


therealharpaljadeja (approved):
Great catch! Thanks!

PR #1350, opened 2024-07-01
Update core contract addresses

pahor167:
Added missing core contract addresses for both mainnet and alfajores.


therealharpaljadeja (approved):
Thank you for the contribution!

PR #1375, opened 2024-07-13
feat: add Dango related contract addresses deployed on Holesky

therealharpaljadeja:


ezdac (approved):
Looks good!

But I noticed that the newly deployed contracts are missing from the Alfajores link,
for example the `CeloDistributionSchedule`  (compare with [the spec](https://github.com/celo-org/specs/blob/main/src/core_contracts.md#dango) )

That should probably be the only one - but it would make sense to go through the list and add the ones to the doc that are missing. 
I don't know if they necessarily should be in the `Alfajores addresses` category since they will only be used in Cel2, but technically they are also deployed there.

That's your decision ;)

therealharpaljadeja:
> Looks good!
> 
> But I noticed that the newly deployed contracts are missing from the Alfajores link, for example the `CeloDistributionSchedule` (compare with [the spec](https://github.com/celo-org/specs/blob/main/src/core_contracts.md#dango) )
> 
> That should probably be the only one - but it would make sense to go through the list and add the ones to the doc that are missing. I don't know if they necessarily should be in the `Alfajores addresses` category since they will only be used in Cel2, but technically they are also deployed on Alfajores.
> 
> That's your decision ;)

Added `CeloDistributionSchedule` to Dango contract addresses and Alfajores too.

PR #1378, opened 2024-07-19
feat: add guide to bridge Celo from Holesky to Dango using Viem

therealharpaljadeja:


karlb (approved):
I didn't try executing the code, but from reading it, it looks fine!

PR #1459, opened 2024-08-28
Add Gas.zip to Token Bridges in Docs

tomkysar:
Added Gas.zip for as native gas reloading option from over 18+ EVM chains & Solana to CELO native


therealharpaljadeja (changes requested):
Can you add it to the bottom of the list and keep the B in Bridges uppercase?

PR #1498, opened 2024-09-26
Update paths

piersy:
Fixes paths such as /path/to/old/datadir/chaindata to be /path/to/old/datadir/celo/chaindata

Changed the name of environment variable OP_GETH_DIR to OP_GETH_DATADIR to make it more clear that it is the datadir.

Avoided use of the path `/celo` inside docker containers for the celo datadir, since it could easily be confused with the child celo dir under the celo datadir (I used /datadir instead).

Updated the paths for the migration to keep all the files under /migration-files

Removed (I think) unnecessary environment variable setting when running op-geth in docker '-e GETH_DATADIR=/celo'

### DIFF
--- a/docs/cel2/l2-operator-guide.md
+++ b/docs/cel2/l2-operator-guide.md
@@ -212,18 +212,18 @@ With snap sync, you can start an L2 node without migrating or downloading the L1
 Also, if you are using snap sync, you will need to `init` the chaindata dir using the provided genesis file. You should not `init` if you are starting your node with a migrated datadir. Run using the container or the binary according to your preference. You can use the following example as a reference.
 
 ```bash
-OP_GETH_DIR=/tmp/alfajores/op-geth
+OP_GETH_DATADIR=/tmp/alfajores/op-geth
 OP_GETH_IMAGE=us-west1-docker.pkg.dev/devopsre/celo-blockchain-public/op-geth:celo8
-mkdir -p ${OP_GETH_DIR}
+mkdir -p ${OP_GETH_DATADIR}
 wget https://storage.googleapis.com/cel2-rollup-files/alfajores/genesis.json --output-document=${OP_NODE_DIR}/genesis.json
 
 docker run -it \
   --name op-geth-init \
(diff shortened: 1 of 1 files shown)


docs/cel2/l2-operator-guide.md
```diff
@@ -368,16 +368,15 @@ make geth
...
 # Copying the jwt secret from the op-node directory
-cp ${OP_NODE_DIR}/jwt.txt ${OP_GETH_DIR}/jwt.txt
+cp ${OP_NODE_DIR}/jwt.txt ${OP_GETH_DATADIR}/jwt.txt
 
 docker run -d \
   --name op-geth \
   --network=host \
-  -v ${OP_GETH_DIR}:/celo \
-  -e GETH_DATADIR=/celo \
+  -v ${OP_GETH_DATADIR}:/datadir \
```
piersy, line 376: @jcortejoso can you verify if this is a legitimate change?

jcortejoso (approved):

docs/cel2/l2-operator-guide.md
jcortejoso: `GETH_DATADIR` helps as you don't need to carry over the `--datadir` flag for any other command. Especially convenient for `geth attach`

jcortejoso:
Also something a noticed is folders under `/tmp` may be problematic for MacOS environments, or at least I could manage to get Docker for Mac volume sharing working with folder under `/tmp` (as /tmp is really a virtual folder under private/tmp).

It may help if we change to some folder under `$HOME` just as safety for some docker corner cases.

sodofi (approved):

docs/cel2/l2-operator-guide.md
piersy: Ahh ok got you, hadn't thought about that.

piersy:
Hi @jcortejoso I was also considering to go for $HOME but for a different reason that if you put files under /tmp you lose them when you restart, which could be problematic for people. Anyway this is merged now, but we can do another PR for that.

PR #1503, opened 2024-09-27
Fixes and clarifications to node operator docs

piersy:
Lots of fixes to node operator docs.

* Invalid flags, op-geth had some flags of op-node
* Incorrect flags given (i.e syncmode=consensus-layer for snap sync)
* Missing `--platform linux/amd64` for docker images that don't have arm support
* Missing backslashes to allow splitting commands over multiple lines
* We were echoing the command to generate the jwt into the file
* Invalid commands given to docker images that have entrypoints set
* Incorrect paths
* Unquoted aterisks breaking execution in shells
* Some clarification and cleanup

### DIFF
--- a/docs/cel2/l2-operator-guide.md
+++ b/docs/cel2/l2-operator-guide.md
@@ -111,7 +111,7 @@ fi
 ```bash
 EIGENDA_IMAGE=us-west1-docker.pkg.dev/devopsre/celo-blockchain-public/eigenda-proxy:v1.4.1
 
-docker run -d \
+docker run --platform linux/amd64 -d \
   --name eigenda-proxy \
   -v ${EIGENDA_KZG_PROXY_DIR}:/data \
   --network=host \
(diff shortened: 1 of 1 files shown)


docs/cel2/l2-operator-guide.md
```diff
@@ -179,7 +179,6 @@ docker run -d \
     --rollup.config=/data/rollup.json \
     --verifier.l1-confs=4 \
     --rpc.addr=127.0.0.1 \
-    --syncmode=<consensus-layer/execution-layer>
```
jcortejoso, line 182: I think `--syncmode` is necessary here:

```
/ # op-node --help | grep -A1 syncmode
    --syncmode value                    (default: consensus-layer)         ($OP_NODE_SYNCMODE)
          Blockchain sync mode (options: consensus-layer, execution-layer)
```

jcortejoso (approved):
Nice! Thanks for looking

docs/cel2/l2-operator-guide.md
piersy: The default is to use consensus-layer, and the instructions before this block (line 161) say to append `--syncmode=execution-layer` if you want to use snap sync. That seems reasonable to me.

PR #1507, opened 2024-10-02
Added Transporter to Celo bridge list

melchermaxwell:


therealharpaljadeja (approved):
Thank you for your contribution!

However @melchermaxwell, note that we are actively migrating Celo Mainnet to an L2, more details about the L2 testnet [here](https://forum.celo.org/t/alfajores-goes-l2/9052).

Would recommend preparing for the migration as soon as possible.

PR #1555, opened 2024-11-04
update to tab headers

sodofi:
1. changed the headers to be 'get started' and 'build with celo'
2. updated more tab to take out repeated links

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -104,12 +104,12 @@ module.exports = {
         },
         {
           to: "build/",
-          label: "Build on Celo",
+          label: "Get Started",
           position: "left",
         },
         {
           to: "developer/",
-          label: "Developers",
+          label: "Build on Celo",
           position: "left",
         },
         {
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -134,20 +138,7 @@ module.exports = {
...
-              label: "Tutorials",
-              target: "_blank",
-            },
-            { to: "showcase", label: "DApps" },
-            {
-              to: "community/guidelines",
-              label: "Community",
-            },
+            { to: "https://www.celopg.eco/ecosystem", label: "dApps", target: "_blank" },
           ],
```
GigaHierz, line 142: please delete these pages as well @sodofi , so they can't be founde/ referenced

docusaurus.config.js
```diff
@@ -104,12 +104,12 @@ module.exports = {
...
           to: "build/",
-          label: "Build on Celo",
+          label: "Get Started",
           position: "left",
         },
         {
           to: "developer/",
-          label: "Developers",
+          label: "Build on Celo",
           position: "left",
```
GigaHierz, line 113: why do we not call this section tools? The top page in this section is called Tools.

docusaurus.config.js
sodofi: I think they should be found/referenced just not from tab navigation

GigaHierz (approved):

PR #1571, opened 2024-11-25
Update evm-tools.md

arshadkazmi42:
The old url was using a github username which was not available on github.com

https://github.com/AgileAlpha

But when we access the repository url https://github.com/AgileAlpha/eth_contract it redirects to https://github.com/zyield/eth_contract
This happens when user change their username.

This PR, updates the correct URL.


sodofi (approved):
Confirmed github URL

PR #1583, opened 2024-12-02
Remove references to Plumo and ultra-light clients

lvpeschke:
Plumo and the ultra-light client are not supported on L2. See https://docs.celo.org/cel2/whats-changed/l1-l2.

https://github.com/celo-org/celo-blockchain-planning/issues/712


aaronmgdr (approved):
i see no issues

PR #1597, opened 2024-12-06
Update the hardfork blocks and releases

piersy:
Removes reference to baklava L2 hardfork block and release where that is set .


karlb (approved):
The https://github.com/celo-org/celo-blockchain/releases/tag/v1.8.7 release does not exist yet. Otherwise the PR is fine.

PR #1638, opened 2025-02-03
Improve builders page

palango:


sidebars.js
```diff
@@ -1262,8 +1262,8 @@ const sidebars = {
     },
     {
       type: "doc",
-      label: "Developers & Builders",
-      id: 'cel2/developers-builders',
+      label: "Builders",
+      id: 'cel2/builders',
```
karlb, line 1266: Will this break existing (external) links? If so, it might be better to just keep the old name.

katen0ra (approved):

PR #1644, opened 2025-02-05
Update operator docs

palango:
For some reason https://github.com/celo-org/docs/pull/1642 results in some build errors, so I refreshed the yarn lockfile here as well.


karlb (approved):
I don't know if you intentionally added the link anchors to all headings, but I assume that is just what the current docusaurus version (or one of its plugins) does. I makes sense to add them, so that they won't change on minor heading text edits.

PR #1653, opened 2025-02-10
Page for SubQuery under the "Indexers" category

bgodlin:


sodofi:
@bgodlin the PR did not build properly.

Can you remerge with main and then run `yarn build` to confirm it will build correctly and then submit a PR again?

PR #1659, opened 2025-02-12
Add note on continuity script

alecps:
Most users who encounter gaps will probably choose to just try re-running with different source dbs until they find one without gaps. Nonetheless, I've added this note here so that the continuity script is mentioned somewhere in our docs in case people get the error and want more information, and also in case they want to know how to run the script without having the full migration script attached to it for whatever reason. In retrospect this might not be essential to have in the docs, so we can skip adding this if it feels superfluous. 

Drive-by changes: 
- I noticed a sentence saying that we don't modify the source db in our migration scripts, which actually isn't true. When we open the db in readwrite mode some `.meta` files are added automatically, which is necessary for op-stack compatibility. I thought it might make sense to remove this sentence entirely to avoid confusion.


piersy (approved):
Great!

PR #1667, opened 2025-02-14
l2: Combine operator guides

palango:


docs/cel2/operators/run-node.md
```diff
@@ -1,10 +1,56 @@
...
+   op-geth`. This will display the last 50 lines of the logs and follow the logs
+   as they are written. In a syncing node, you would expect to see lines of the
+   form `Syncing beacon headers  downloaded=...` where the downloaded number is
+   increasing and later lines such as `"Syncing: chain download in
+   progress","synced":"21.07%"` where the percentage is increasing. Once the
+   percentage reaches 100%, the node should be synced.
+6. At this point, you should be able to validate the progression of the node by
+   fetching the current block number via the RPC API and seeing that it is
+   increasing. (Note that until fully synced, the RPC API will return 0 for the
+   head block number).
```
karlb, line 43: This part is mostly redundant with the repo description, with the usual difficulties of keeping both up to date and consistent. But I assume that is an intentional tradeoff in this case.

docs/cel2/operators/run-node.md
```diff
@@ -35,12 +81,12 @@ Docker images are the easiest way to run an Celo node, but you can always build
...
-  - [eigenda-proxy](https://ghcr.io/layr-labs/eigenda-proxy:v1.4.1)
+  - [op-geth](https://us-west1-docker.pkg.dev/devopsre/celo-blockchain-public/op-geth:celo-v2.0.0-rc3)
+  - [op-node](https://us-west1-docker.pkg.dev/devopsre/celo-blockchain-public/op-node:celo-v2.0.0-rc3)
+  - [eigenda-proxy](https://ghcr.io/layr-labs/eigenda-proxy:v1.4.3)
 
-## Common problems
+### Common problems
 
-### Transactions are not being executed when submitted to a node
+#### Transactions are not being executed when submitted to a node
```
karlb, line 90: Should this common problems section really only relate to the "build from source"? If it is meant to be more general, the heading change should not be done like this.

palango: The problem would not occur with our repo, that's why I added it under the building form source section.

PR #1680, opened 2025-02-19
Baklava docs update

pahor167:


piersy:
Hey @pahor167, I'm not really sure what marek meant when he asked for this the run celo node page to be updated with contract information. But this addition doesn't really make any sense to me, it's not clear what it is, a bunch of config parameters, but for what, where should they be used? They are not actually needed for running a node so it doesn't seem to make sense to have them in this page. But at the very least there needs to be some explanation of what they are, and how they should be used.

PR #1682, opened 2025-02-19
Link to RPC diff description in specs

karlb:
The relevant information is added in https://github.com/celo-org/specs/pull/89

Closes https://github.com/celo-org/specs/issues/84?reload=1


karlb:
While I was waiting for https://github.com/celo-org/specs/pull/89, this has already been changed.

PR #1693, opened 2025-02-20
Add links to baklava assets in the node run docs

piersy:
As in the title, adds links to all the baklava assets and provides the ips of the baklava nodes.


palango (approved):
Thanks!

Issue #1704, opened 2025-02-22
Redirect for node operator instructions


lvpeschke:
We should make sure that previous links we shared with node operators keep working.

Redirect https://docs.celo.org/cel2/l2-operator-guide to https://docs.celo.org/cel2/operators/overview.

PR #1713, opened 2025-02-26
l2: Add redirect

palango:
Closes https://github.com/celo-org/docs/issues/1704

Docs: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects#ex-config

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -31,16 +31,21 @@ module.exports = {
     // },
   ],
   plugins: [
+    [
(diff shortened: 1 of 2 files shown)


palango:
It doesn't work for some reason 🤔

lvpeschke:
> It doesn't work for some reason 🤔

The preview works for me: https://deploy-preview-1713--celo-docs.netlify.app/cel2/l2-operator-guide

palango:
Oh, not sure what I was checking...

PR #1728, opened 2025-03-07
Uniswap contract addresses

atejada:
Add Uniswap contract addresses to the Celo Alfajores list


sodofi:
@GigaHierz @viral-sangani can you review? what's your opinion on where uniswap contracts should go?

sodofi:
Looking at this PR https://github.com/celo-org/docs/pull/1729 this seems like the best spot these CAs should go.

atejada:
Sorry Sophia, I was supposed to talk with you about this...I did this before doing the other, so you made the right call 😌

PR #1741, opened 2025-03-14
Update Token-Addresses

atejada:
Add Tokens to Celo and Alfajores

I check the first 50 pages of transactions from https://celoscan.io/tokentxns and https://alfajores.celoscan.io/tokentxns


sodofi:
Can you update so you are only including Mento's official tokens. The ones listed here: https://reserve.mento.org/ @atejada

atejada:
@sodofi Fixes are done ✅ Not sure if Alfajores Token should be revised as well 🤔

sodofi (approved):

PR #1762, opened 2025-03-18
L1->L2: how to decrypt keystore

karlb:
Closes https://github.com/celo-org/celo-blockchain-planning/issues/938


palango (approved):
Thanks!

PR #1766, opened 2025-03-18
fix: Enhance list styling in custom CSS to restore default behaviors …

### LINKED ISSUE #1763: List symbols not shown
Image

See screenshot.

The numerals for list items are not shown. This might be because of `list-style: none;`

viral-sangani:
Removed by Tailwind v4. Added styles for unordered and ordered lists, including nested lists for better visual hierarchy.

Resolves #1763

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -18,6 +18,36 @@
   ::file-selector-button {
     border-color: var(--color-gray-200, currentColor);
   }
+  
+  /* Restore list styling that Tailwind v4 removes by default */
+  ul {
+    list-style-type: disc;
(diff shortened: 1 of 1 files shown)


palango (approved):

palango:
I'm a bit surprised that we need to manually define those, but it fixes the issue. 

Isn't there some default docusaurus style that we can derive from?

viral-sangani:
Hey @palango , yes it was initially handled by Tailwind, but in the last Tailwind v4 update, they removed few features. Most probably there will be some plugin that can add these styles. Will look into it.

PR #1770, opened 2025-03-19
Create Fund Your Project Page

atejada:
Create a new page after 'integrate with celo' called 'Fund your project.'


sodofi (approved):
Try to ensure that the style matches the styling on other pages.

Ex: Only lines at the header, number lists don't also include a dash, etc. See other pages for examples.

PR #1771, opened 2025-03-19
Fix Support Page

atejada:
Update page to 'Get Support' and fix styling. Take out Celo Signal Mailing list and update to 'DevDesk Mailing List'

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -726,7 +726,7 @@ const sidebars = {
     },
     {
       type: "doc",
-      label: "Support",
+      label: "Get ",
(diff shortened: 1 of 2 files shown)


sidebars.js
```diff
@@ -726,7 +726,7 @@ const sidebars = {
     },
     {
       type: "doc",
-      label: "Support",
+      label: "Get ",
```
palango, line 729:
```suggestion
      label: "Get Support",
```

sodofi (approved):

PR #1778, opened 2025-03-20
Update migrate docs

piersy:
Update the docs for migration to cover both mainnet and testnet migrations which differ slightly now, since the mainnet migration requires an extra flag.


palango (approved):
Fancy solution with the tabs. I like that!

PR #1782, opened 2025-03-20
Integrate with Celo Overhaul

atejada:
Edit the old 'Integrate with Celo' to match the new integration section written here: https://docs-cel2.netlify.app/build/integrations/


atejada:
All changes have been added, so nothing to do here 🫡

PR #1811, opened 2025-03-25
Update run-node.md

alvarof2:


piersy (approved):
Looks good if you say so @alvarof2!

PR #1835, opened 2025-03-27
Add redirects 2

GigaHierz:
- redirect 2nd try


static/_redirects
```diff
@@ -691,4 +691,6 @@
 /protocol/socialconnect https://self.xyz/
 /contract-addresses /contracts/core-contracts 301
 /token-addresses /contracts/token-contracts 301
-/developer/build-on-minipay /developer/build/build-on-minipay
\ No newline at end of file
+/developer/build-on-minipay /developer/build/build-on-minipay/overview
+/developer/build/build-on-minipay /developer/build/build-on-minipay/overview
```
lvpeschke, line 695: Tested on the preview instance and these don't work :/

lvpeschke: Should it link to /build/build-on-minipay/overview?

GigaHierz:
@lvpeschke added all redirects for the big refactoring here

palango:
I got some more warnings about links that worked before here: https://github.com/celo-org/specs/issues/112

static/_redirects
```diff
@@ -681,14 +680,100 @@
...
+/protocol/pos/epoch-rewards-community-fund /what-is-celo/using-celo/protocol/epoch-rewards/community-fund
+/protocol/pos/epoch-rewards-carbon-offsetting-fund /what-is-celo/using-celo/protocol/epoch-rewards/carbon-offsetting-fund
+/protocol/consensus/index /what-is-celo/using-celo/protocol/consensus
+/protocol/consensus/validator-set-differences /what-is-celo/about-celo-l1/protocol/consensus/validator-set-differences
+/protocol/consensus/locating-nodes /what-is-celo/about-celo-l1/protocol/consensus/locating-nodes
+/protocol/consensus/ultralight-sync /what-is-celo/using-celo/protocol/consensus
+/protocol/identity/index https://self.xyz/
+/protocol/randomness /what-is-celo/about-celo-l1/protocol/randomness
+/protocol/contracts/add-contract /what-is-celo/about-celo-l1/protocol/contracts/add-contract
+/protocol/transaction/erc20-transaction-fees /what-is-celo/using-celo/protocol/transaction
```
lvpeschke, line 741: /what-is-celo/using-celo/protocol/transaction doesn't work

lvpeschke: There is another redirect for the same link below, so I recommend removing the line

docs/cli/commands.md
lvpeschke, line 1: This will get overwritten by automation again.

docs/cli/plugins.md
lvpeschke, line 1: This will get overwritten by automation again.

docs/what-is-celo/using-celo/protocol/transaction/overview.md
```diff
@@ -3,9 +3,9 @@ title: Transactions on Celo
 description: Introduction to transactions on Celo.
 ---
 
-# Celo Token
+# Transactions on Celo
 
-Introduction to transactions on Celo, gas fees 
+Introduction to transactions on Celo.
 
 :::warning
```
lvpeschke, line 10: I suggest removing this warning, as this page takes the L2 into account.

lvpeschke (changes requested):
Thanks for the additions!
The redirects file will need additional cleanup, but these added links are a step into the right direction.

lvpeschke (approved):

PR #1840, opened 2025-04-03
chore: sort _redirects and remove duplicates

lvpeschke:
- Sort file for easier management
- Add link to framework documentation
- Remove duplicates: identical entries and redirects with different targets


palango (approved):

lvpeschke:
> If you have a command handy, we could add a check for duplicates to CI

Nothing I have handy, but would be good to add: https://github.com/celo-org/docs/issues/1846

PR #1860, opened 2025-04-16
Validator payment command

pahor167:


pahor167:
> I can't comment on the validity of commands, but overall this reads really fluent and seems helpful.
> 
> Maybe some more context (links to docs, commands) on the commission rate would be nice

Added command links

PR #1868, opened 2025-04-23
l2: Add Isthmus page

palango:
Add page with Isthmus information.

Side questions: Do we want to keep the information bar and the notice for the migration, or should we remove those?

Resolves https://github.com/celo-org/celo-blockchain-planning/issues/1012

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -1208,6 +1208,11 @@ const celoL2Sidebar = [
     collapsed: false,
     items: [
       {
+        type: "doc",
+        label: "Isthmus Upgrade",
+        id: "cel2/notices/isthmus-upgrade",
+      },
+      {
         type: "category",
         label: "Celo L2 Migration",
         link: {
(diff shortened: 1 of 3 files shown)


sidebars.js
```diff
@@ -1225,7 +1230,7 @@ const celoL2Sidebar = [
       },
       {
         type: "doc",
-        label: "Pectra Upgrade",
+        label: "L1 Pectra Upgrade",
```
lvpeschke, line 1233: Could you highlight here that for Mainnet, running the migration version is enough? It's in the callout, but not the first thing that catches a reader's attention; I'd add a title for Mainnet.

lvpeschke (approved):
> Do we want to keep the information bar 

I would keep, not a strong opinion

> and the notice for the migration, or should we remove those?

Ideally remove, but the information from that page needs to be worked into "Migrating an L1 Node", which currently refers to that one. There is also no other place where min versions for each network are mentioned.

PR #1878, opened 2025-04-30
Rename script usages to match updated package.json in protocol.

Mc01:
Companion PR to https://github.com/celo-org/celo-monorepo/pull/11369


lvpeschke (approved):

soloseng (approved):
🚀

martinvol (approved):
I didn't verify if there are more files that need to be updated, but the content of this PR looks good!

PR #1881, opened 2025-05-06
Bridging Native ETH to Celo

pahor167:


docs/cel2/notices/native-ETH-bridging.md
```diff
@@ -0,0 +1,47 @@
+# SuperBridgeETHWrapper: Bridging Native ETH to Celo
+
+This document provides a minimal overview of the `SuperBridgeETHWrapper.sol` contract and its deployed addresses.
+
+## Purpose
+
+The `SuperBridgeETHWrapper.sol` contract allows users to bridge native ETH from an L1 network (like Ethereum or Holesky) to the Celo L2 network, where it arrives as WETH.
+
+> **Disclaimer:** When using Superbridge, this entire process is abstracted away from the user.
```
karlb, line 9: I would consider this to be a "note" rather than a disclaimer.

lvpeschke:
@pahor167 you'll also need to add the page to https://github.com/celo-org/docs/blob/main/sidebars.js. I would also recommend against notices, which are more timing related as opposed to general guides and references. Maybe a section under https://docs.celo.org/what-is-celo/using-celo/bridges and/or https://docs.celo.org/cel2/guides/bridging-celo-from-l1-to-l2 instead?

karlb (approved):

soloseng (approved):

lvpeschke:
@pahor167 could you address the review comments and merge?

PR #1886, opened 2025-05-12
Update fee abstraction guide

karlb:
* The token addresses were not the mainnet addresses. Since no other network was mentioned and the example at the end of the page refers to mainnet, I assume they should be mainnet addresses.
* Update addresses in CLI example
* FeeCurrencyDirectory instead of old whitelist

### DIFF
--- a/docs/cel2/guides/fee-abstraction.md
+++ b/docs/cel2/guides/fee-abstraction.md
@@ -4,7 +4,7 @@ title: Fee Abstraction on Celo L2
 
 ## Fee Abstraction Addresses
 
-Celo allows paying gas fees in currencies other than the native currency. The tokens that can be used to pay gas fees are controlled via governance and the list of tokens allowed is maintained in `FeeCurrencyWhitelist.sol`.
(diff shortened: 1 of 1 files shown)


palango (approved):
🎉

karlb:
Alfajores or mainnet addresses in the table, what do you think? If we go with the mainnet, I should at least change 
```bash
celocli network:whitelist --node https://alfajores-forno.celo-testnet.org
```
to also point to the mainnet.

palango:
> Alfajores or mainnet addresses in the table, what do you think? If we go with the mainnet, I should at least change
> 
> ```shell
> celocli network:whitelist --node https://alfajores-forno.celo-testnet.org
> ```
> 
> to also point to the mainnet.

I think showing the addresses and example for Mainnet makes more sense.

Optionally we can add a small section that shows how to get the values for the testnets?

karlb:
I added the CLI commands for both alfajores and testnet and added an explicity "mainnet" to the table column labels to avoid ambiguity.

PR #1907, opened 2025-05-28
isthmus: adjust activation

palango:


lvpeschke (approved):

gastonponti (changes requested):
Wait. Those timestamps are in May

gastonponti (approved):
Timestamps updated in the last commit

PR #1912, opened 2025-06-02
tokens: Update bridgeable tokens

palango:
Closes https://github.com/celo-org/celo-blockchain-planning/issues/971


docs/what-is-celo/using-celo/bridged_tokens/tokens.md
```diff
@@ -1,7 +1,24 @@
 | Symbol | Token Name | Contract Addresses |
 | ------ | ---------- | ------------------ |
+| 1INCH | 1INCH Token (Celo native bridge) | L1: [0x111111111117dC0aa78b770fA6A738034120C302](https://etherscan.io/token/0x111111111117dC0aa78b770fA6A738034120C302)<br />L2: [0x28ba8d26f5f6710f42170ee545a0c953ca4997b9](https://celoscan.io/token/0x28ba8d26f5f6710f42170ee545a0c953ca4997b9) |
```
lvpeschke, line 3: Why add the L2 full name to some but not all?
For consistency, I'd keep either the L1 or the L2 name, or both, but not a mix

lvpeschke (approved):

docs/what-is-celo/using-celo/bridged_tokens/tokens.md
karlb: The data is inconsistent in the data source. The L2 name is only given when there is an explicit override in https://github.com/ethereum-optimism/ethereum-optimism.github.io/tree/master/data. OP's validation should take care that there always is an override when the L2 token name differs from the L1 name, but apparently this is nothing we can rely on.

I've removed the ` (Celo native bridge)` as it resolved the problem in this case and improves the readability.

PR #1954, opened 2025-07-04
eclair: Add info

palango:


ezdac (approved):
Looks  good - what about funds? Should we link a contact to give out funds?

PR #1961, opened 2025-07-14
Add block numbers for isthmus hardfork

piersy:
It's useful to be able to reference the block numbers directly rather than having to find the block.


palango:
Please add links to the explorer on the block numbers.

palango (approved):

PR #1982, opened 2025-07-24
Celo L1 validator refactor

lvpeschke:
Update warnings and notices on validator documentation.
Move Celo Signal page out of L1.


pahor167 (approved):

sodofi:
> @sodofi can you confirm if Celo Signal is still in use? Removed for now.

Hey @lvpeschke we can remove Celo Signal

PR #1991, opened 2025-07-28
Add EigenDA , native Bridging

GigaHierz:
1. Adding EigenDA
2. Adding Native Bridging

Open questions here, related to the [spec](https://specs.celo.org/bridge.html):

    - Should we use Superchain docs for WETH bridging? Harpal did the docs for this, but in the spec, you mention using optimism sdk: 
      - https://docs.celo.org/what-is-celo/using-celo/native-ETH-bridging 
      - https://docs.optimism.io/stack/transactions/deposit-flow
    - Intro:
        - Does CELO have different addresses on Ethereum than Celo? 
        - Don’t understand that sentence: The L2 version is the native token on the L2 Celo chain, preserving the balances from the Celo L1 chain. How does it preserve the balances from the Celo L1 chain? 
        - Do we have an address for CeloDistributionSchedule?
    - Bridging tokens
        - Bridging CELO will be different from ERC20 tokens, even if CELO is also an ERC20 token?
        - Where is the fee abstraction happening? During the bridging? Only on Celo? Is this specific to bringing or all Fee Currency? 
        - What are the explorer tools for checking bridging?

### DIFF
--- a/docs/what-is-celo/using-celo/eigenda.md
+++ b/docs/what-is-celo/using-celo/eigenda.md
@@ -1,0 +1,86 @@
+---
+title: EigenDA
+description: Scaling Data Availability for the Next Generation of Web3 Apps
+---
+
+
+# EigenDA on Celo
+
+Celo is now an Ethereum Layer 2 (L2) designed for fast, low-cost, mobile-first transactions. As part of this upgrade, Celo integrates **EigenDA**, a high-throughput, decentralized data availability (DA) layer built on **EigenLayer**.
+
+This integration,[first proposed on the Celo Forum](https://forum.celo.org/t/proposal-eigenda-integration/6382),solves one of the core challenges facing modern L2s: how to keep transaction fees low and predictable without compromising security or scalability.
+
+---
+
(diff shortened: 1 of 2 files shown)


viral-sangani:
@claude Can review this PR also?

GigaHierz:
Closing as superseded — this one can't be merged as-is and its content already lives on main in better form.

It dates to July 2025 and targets the old Docusaurus layout (`docs/…` paths + `sidebars.js`), which no longer exists after the migration to Mintlify (`docs.json` + `.mdx`). Both topics it introduced now have dedicated, more complete pages on `main`:

- **Native bridging** → `specs/native-bridge.mdx`, `home/bridged-tokens/native-ETH-bridging.mdx`, `home/bridged-tokens/bridges.mdx`.
- **EigenDA** → `specs/eigenda.mdx`, plus the EigenDA upgrade notices under `infra-partners/notices/archive/`.

Nothing here needs porting. Leaving the branch (`add-cel2-infra-intro`) in place in case any wording is worth a look; no follow-up issue needed.

PR #2021, opened 2025-08-19
Update to-celo-cli.md

lvpeschke:
The page had a mismatch between title and content for connecting to Alfajores vs Mainnet. Fix that and update to Celo Sepolia with tabs, to highlight that it's one or the other, not both.


docs/wallet/ledger/to-celo-cli.md
```diff
@@ -27,33 +27,33 @@ And if not, upgrade by running the same command as above.
...
+    ```
+  </TabItem>
+  <TabItem value="mainnet" label="Celo Mainnet" default>
+    Configure the Celo CLI so that it uses a cLabs node on the Celo network.
+
+    ```bash
+    celocli config:set --node https://forno.celo.org/
+    ```
+  </TabItem>
+</Tabs>
```
GigaHierz, line 56: I would set mainnet as first tab and sepolia as second tab

PR #2031, opened 2025-08-29
Update index.md

sodofi:


GigaHierz:
hi @sodofi ,

I believe this should be reviewed by Kate, as she was in the hackmamba call. It looks incomplete to me based on the conversations we had in the call.

viral-sangani:
@sodofi can you move this changes mintlify compatible file.

PR #2047, opened 2025-09-30
ci: add GitHub workflows for documentation validation and Mintlify checks

viral-sangani:
- Broken Links Fixed (92 total)
- Script Improvements
- Added CICD to check valid docs

### DIFF
--- a/.github/workflows/docs-validation.yml
+++ b/.github/workflows/docs-validation.yml
@@ -1,0 +1,208 @@
+name: Documentation Validation
+
+on:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    branches:
+      - main
+      - master
+  workflow_dispatch:
+
+jobs:
+  validate-docs:
+    name: Check for Errors and Broken Links
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      - name: Install Mintlify CLI
+        run: npm install -g mintlify
+
+      - name: Check for broken links
+        id: broken-links
+        continue-on-error: true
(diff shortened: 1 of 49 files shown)


.github/workflows/docs-validation.yml
```diff
@@ -0,0 +1,208 @@
...
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
```
palango, line 26: There's a newer version available.

.github/workflows/docs-validation.yml
```diff
@@ -0,0 +1,208 @@
...
+        with:
+          node-version: '20'
```
palango, line 28: Latest LTS is 22, maybe we should use that?

.github/workflows/docs-validation.yml
```diff
@@ -0,0 +1,208 @@
...
+            echo "✅ No broken links found!" >> $GITHUB_STEP_SUMMARY
+          else
+            echo "❌ Broken links detected:" >> $GITHUB_STEP_SUMMARY
+            echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
+            cat broken-links-output.txt >> $GITHUB_STEP_SUMMARY
+            echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
+            exit 1
+          fi
+
+      - name: Validate documentation structure
```
palango, line 52: This seems duplicated with `./scripts/validate-docs.sh`. Isn't there some `mint` commands we can use instead of re-implementing all of those checks?

viral-sangani: Yes, my bad. I agree we have validate docs. I have updated the CICD to use that script, and mintlify do not give any tools in their CLI to do these validations

.github/workflows/docs-validation.yml
viral-sangani: Updated

PR #2071, opened 2025-10-17
replace broken link "about-celo"

GigaHierz:


palango:
When I try in the test deployment the link still doesn't work.

PR #2092, opened 2025-11-05
jello: Add Celo Sepolia activation time

palango:


infra-partners/notices/jello-upgrade.mdx
```diff
@@ -7,7 +7,7 @@ This page outlines changes related to the Jello Upgrade for node operators.
 <Note>
 This page will be kept updated with key information about the upgrade. As this upgrade is activated via a smart contract switch, no concrete activation time can be given.
 
-- Celo Sepolia testnet activation is planned for end-October, 2025.
+- Celo Sepolia testnet activation was executed executed on Wed, Oct 5, 2025.
 - **Mainnet** activation is planned for mid-November, 2025.
```
piersy: Should we add a block or some other kind of reference, like the first L2 block that will be part of a type 42 fault dispute game?

PR #2173, opened 2026-04-15
Add notice for Req/Res CL P2P sync deprecation

palango:
Celo is following Optimism on deprecating the op-node request-response consensus-layer P2P sync client in favor of execution-layer syncing.


piersy (approved):
All good 👍

PR #2193, opened 2026-06-09
docs: add wagmi v2 + viem v2 MiniPay integration example with hydration fix

2TheMoom:
Adds a complete wagmi v2 + viem v2 integration example for MiniPay including:

- createConfig with Celo chain
- useMiniPay detection hook with auto-connect
- Hide connect button pattern
- React hydration error fix for Next.js 14 apps
- wagmi v1 → v2 migration table

Based on a working implementation built for Celo Proof of Ship June 2026.
Live demo: https://celosense.vercel.app

Built and tested by Abu Olumi

### DIFF
--- a/build-on-celo/build-on-minipay/code-library.mdx
+++ b/build-on-celo/build-on-minipay/code-library.mdx
@@ -406,3 +406,146 @@ async function requestTransfer(tokenAddress, transferValue, tokenDecimals, recei
 ```
 
 {/* prettier-ignore-end */}
+
+
+## MiniPay with wagmi v2 + viem v2
+
+If you are building with **wagmi v2** and **viem v2** (the current standard for Next.js 14 apps), here is the complete integration pattern for MiniPay detection and auto-connect.
+
+### Install dependencies
+
+```bash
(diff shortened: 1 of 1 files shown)


GigaHierz:
Thanks for this! The wagmi v2 patterns here are genuinely useful, so I've incorporated them — with a few corrections — into the MiniPay quickstart instead, where the wallet setup naturally lives: #2195.

A couple of reasons for the different placement:
- The target page (`code-library.mdx`) is already viem v2 throughout, so it wasn't actually outdated. The stale wagmi **v1** content was in `quickstart.mdx`, which #2195 updates.
- The "hide connect button" / auto-connect material overlapped with what the quickstart already covered, so consolidating avoids two places teaching the same thing.

Fixes applied while folding it in: the connector now matches between `createConfig` and `connect()` (the original registered bare `injected()` but connected with `injected({ target: "metaMask" })`), the imports/chains source are consistent, and a couple of copy-paste leftovers were dropped.

Closing this in favor of #2195 — thanks again for the contribution!

PR #2199, opened 2026-06-17
Add Openfort to Wallet Infrastructure

joalavedra:
## Description

Adds [Openfort](https://www.openfort.io/) to the **Wallet Infrastructure** section of `tooling/wallets/index.mdx`, alongside Privy, Para, Portal, and thirdweb.

Openfort is wallet infrastructure for developers, offering non-custodial embedded wallets with social login, programmable backend (server) wallets, and account abstraction with paymasters for gasless transactions and paying fees in stablecoins — all relevant to building on Celo.

### Entry added

- Homepage: https://www.openfort.io/
- Docs: https://www.openfort.io/docs
- Source Code: https://github.com/openfort-xyz

## Changes

- `tooling/wallets/index.mdx` — new Openfort entry under "Wallet Infrastructure" matching the existing format.

PR #2203, opened 2026-06-20
docs: use X-PAYMENT header consistently in x402 Next.js example

2TheMoom:
The PAYMENT-SIGNATURE fallback is not part of the x402 spec.  The protocol uses only X-PAYMENT. Simplified to one consistent header.


GigaHierz:
Superseded by #2204, which combines this change with #2199. Closing in favor of the combined PR.

PR #2224, opened 2026-07-21
docs: migrate specs.celo.org into a new Specs tab

palango:
## What

Moves the Celo L2 specification from specs.celo.org (mdBook, [celo-org/specs](https://github.com/celo-org/specs)) into this repo:

- All 16 spec pages converted to MDX under `specs/`, faithful to the source (doctoc TOCs stripped, H1s moved to frontmatter, links rewritten to internal paths).
- New **Specs** tab with sidebar sections *Specification* (Introduction, Deployments incl. L1 verification, Token Duality, Transaction Fees, Fee Abstraction, Transaction Types, Native Bridge, EigenDA, Finality) and *Protocol Upgrades* (Jovian, Jello, Ice Cream, Isthmus, L2 Migration incl. Contract Updates).
- The external-link stub `infra-partners/specs` is removed; `/infra-partners/specs` redirects to `/specs`.
- 24 links in live sections that pointed at specs.celo.org now use internal paths (`_deprecated/` untouched).
- `/specs/` assigned to @celo-org/blockchain in CODEOWNERS.
- Small content updates: completed events (L2 migration, Ice Cream, Isthmus) described in past tense; Jovian now targets Mainnet and Celo Sepolia, matching its own timeline table.

## Merge order

Merge and deploy this PR **first**. A follow-up PR in celo-org/specs then replaces the mdBook site with redirect stubs pointing at these pages, so old specs.celo.org URLs keep working (fragments preserved).

## Open items for @celo-org/blockchain

- `specs/transaction-fees.mdx` still contains the old note that the mainnet FeeHandler behavior is unconfirmed "as of 2024-11-15" (left verbatim on purpose). Current on-chain config: burn and carbon fractions are zero, 100% of fees go to a cLabs-named beneficiary.
- Relatedly, `home/protocol/transactions/overview.mdx` claims 10% of transaction fees go to the carbon offset fund, which does not match the current on-chain carbon fraction (zero). Not touched in this PR.
- The orphaned specs pages (`networks/*/network.md` address tables, `block_time_and_gas_limit.md`) were deliberately not migrated; their content is covered by `tooling/contracts/*` and the legacy transition docs.

## Follow-ups

- Content overlap between the migrated spec pages and the existing guides (`specs/token-duality` vs `home/protocol/celo-token`, `specs/fee-abstraction` vs `build-on-celo/fee-abstraction/*`) is tracked in #2227. Deferred so this PR stays a pure move, and so the trim happens once the celo-org/specs stub mapping pins down which heading fragments have to survive.

### DIFF
--- a/specs/fee-abstraction.mdx
+++ b/specs/fee-abstraction.mdx
@@ -1,0 +1,111 @@
+---
+title: "Fee Abstraction: Paying Gas With ERC20 Tokens"
+sidebarTitle: "Fee Abstraction"
+---
+
+Fee Abstraction is a Celo feature that allows users to send transactions without spending any native CELO tokens. Instead, ERC20 tokens are used to pay for the transaction's gas cost.
+
+## As a User
+
+### Sending Fee Abstraction Transactions
+
+To pay gas via Fee Abstraction, the gas token must be specified in a [CIP-64](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0064.md) transaction. This tx type has an additional `feeCurrency` field that specifies which token is used for the `maxFeePerGas` and `maxPriorityFeePerGas` fields. See [CIP-64](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0064.md) for details.
+
+Client library support is [described in docs.celo.org](/home/protocol/transactions/transaction-types#client-library-support).
+
+### JSON-RPC Changes
+
+To make it easier to reliably send Fee Abstraction transactions, the JSON-RPC interface has been extended in two ways:
+
+- `eth_estimateGas` takes an optional `feeCurrency` parameter to get Fee Abstraction specific gas estimates (the gas price for the fee token transfers depends on the token)
+- `eth_gasPrice` and `eth_maxPriorityFeePerGas` take an optional `feeCurrency` parameter for getting CIP-64 gas prices.
+
+### The Fee Currency Directory
+
+Only tokens that have been registered in the `FeeCurrencyDirectory` contract using Celo's governance process can be used for Fee Abstraction. The directory entry shows that the token is trustworthy and contains additional information about the token:
+
+- The token address
+- An oracle address, conforming to the [`IOracle`](https://github.com/celo-org/celo-monorepo/blob/2ace1d08b9da5d2618e945eef7663cc385917a2d/packages/protocol/contracts-0.8/common/interfaces/IOracle.sol) interface, to get the current exchange rate
+- The amount of intrinsic gas for Fee Abstraction gas calculation (see ["Gas Calculation for Fee Abstraction Transactions"](#gas-calculation-for-fee-abstraction-transactions))
+
+You can run the `getCurrencies()` method on the `FeeCurrencyDirectory` contract to get a list of all registered tokens and use `getExchangeRate(token address)` to fetch the current exchange rate for a token. The `IOracle` interface is only intended for parties providing Fee Abstraction tokens, not for users or client libraries.
+
+### Tokens using FeeCurrencyAdapter
+
+Tokens with a low number of `decimals` (USDC and USDT) would cause problems during the gas calculation, because the cost per gas can only be accurately given in fractions and not integers. To avoid this issue, the tokens are not directly added to the `FeeCurrencyDirectory`. Instead, a `FeeCurrencyAdapter` contracts wraps these tokens, calculates the costs per gas in a higher accuracy and only converts it back to the lower accuracy once the total gas costs for a transaction is determined.
+
+This only really impacts the user in a single way: Instead of using the USDC and USDT token addresses, you have to pass their `FeeCurrencyAdapter` addresses in the `feeCurrency` field for CIP-64 txs.
+
+## For Token Authors
+
+### Requirements for Registering a Fee Abstraction Token
+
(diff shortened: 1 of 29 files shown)


specs/upgrades/jovian.mdx
```diff
@@ -0,0 +1,51 @@
+---
+title: "Jovian Upgrade"
+sidebarTitle: "Jovian"
+---
+
+<Warning>
+The Celo Jovian hardfork is still being worked on. This information is preliminary.
```
karlb: Jovian is already live, so this should be updated. Not a problem specific to the migration.

specs/native-bridge.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: "Native Bridge"
+---
+
+With the L2 migration, the Celo blockchain gained a native bridge to Ethereum based on [OP Stack's Standard Bridge](https://docs.optimism.io/app-developers/guides/bridging/standard-bridge). Specifically we use Custom Gas Token feature. This page describes the process of bridging assets between L1 and L2.
+
+The Celo token now exists in both L1 and L2 versions. The L1 version is a standard ERC20 token with a total supply of 1 billion, fully minted to the `OptimismPortal` smart contract, which is part of the bridge (this setup allows any Celo token holder on L2 to bridge their tokens to L1). The L2 version is the native token on the L2 Celo chain, preserving the balances from the Celo L1 chain. Tokens that have not yet been minted on the Celo L1 chain, such as tokens for Community Fund, are now minted to the `CeloDistributionSchedule`, which manages further distribution.
```
karlb: CeloDistributionSchedule is now called CeloUnreleasedTreasury IIRC.

specs/smart-contract-updates-from-l1.mdx
```diff
@@ -0,0 +1,164 @@
+---
+title: "Smart Contract Updates From L1"
+sidebarTitle: "Contract Updates"
+---
+
+Smart contract changes that will be deployed to Celo before the transition can be seen in this diff: [https://github.com/celo-org/celo-monorepo/pull/11035/files](https://github.com/celo-org/celo-monorepo/pull/11035/files)
```
karlb:
```suggestion
Smart contract changes that have been deployed to Celo before the transition can be seen in this diff: [https://github.com/celo-org/celo-monorepo/pull/11035/files](https://github.com/celo-org/celo-monorepo/pull/11035/files)
```

More issues like this in the same file where we talk about future events that are already in the past.

specs/token-duality.mdx
karlb, line 1: This mostly duplicates home/protocol/celo-token.mdx

specs/fee-abstraction.mdx
karlb, line 1: This has a lot of overlap with `build-on-celo/fee-abstraction`. We might want to deduplicate that somehow.

karlb (commented):
Some outdated text that is unrelated to the migration itself and two cases of redundant content.

specs/upgrades/jovian.mdx
palango: Fixed in d902b74. Took the activation timestamps from `infra-partners/notices/archive/jovian-upgrade.mdx` (Celo Sepolia `1773749037`, Mainnet `1774958788`) and dropped the Block Height column, since no block heights are published for this upgrade.

specs/native-bridge.mdx
palango: Fixed in 98e2a67. It was the only remaining `CeloDistributionSchedule` in the repo.

specs/smart-contract-updates-from-l1.mdx
palango: Applied in efb05ec, along with the other cases in the file: the Governance Hotfix note about the pre-transition setup, the FeeHandler section (Contract Release 12 has since been deployed), both slashers under Slashing, the precompile-revert sentence, and the Transition section.

Two I left alone:

- "They still remain in the Celo Registry, but are scheduled for deletion after the L2 fully activates" — I could not confirm whether those registry entries were actually removed, so I did not want to guess.
- The bot sentence now reads "cLabs runs a bot", replacing "cLabs is intended to run a bot". Epochs are clearly being processed, but nothing in the docs states who operates that bot. Happy to hedge it back if that is wrong.

specs/token-duality.mdx
palango: Agreed, the conceptual half is the same content twice. Tracked in #2227 rather than done here, to keep this PR a pure move.

One asymmetry worth noting for whoever picks it up: this page also documents the precompile itself (address, gas cost, ABI parameters, Jovian address warming), none of which is in `home/protocol/celo-token.mdx`. So the merge should go in that direction, not the reverse.

specs/fee-abstraction.mdx
palango: Agreed on the overlap, also tracked in #2227.

The duplicated part is the "For Token Authors" section, which `build-on-celo/fee-abstraction/add-fee-currency.mdx` covers with more detail. The "Inside the Celo Blockchain" half has no home elsewhere, so I would keep that here.

The reason I did not do it in this PR: the celo-org/specs stub PR is meant to preserve URL fragments, and the headings on these pages are its redirect targets. Deleting sections breaks those deep links, and `mintlify broken-links` does not check anchors, so nothing would catch it. Easier to do once the stub mapping exists.

karlb (approved):

Issue #2253, opened 2026-08-21
task: Orphaned-pages audit — re-nav or delete the 22 pages on disk that are missing from navigation


GigaHierz:
Part of #2266.

### What needs to be done?

25 tracked `.mdx` files are on disk but not in `docs.json` navigation at `bdf40b37` (script: string leaves of `navigation` vs `find . -name '*.mdx'` excluding `_deprecated`, `snippets`, `node_modules`). 22 are in scope here; the 3 `build-on-celo/build-with-thirdweb/*` pages belong to #2255.

The worst case is the **entire dev-environment-setup group**, which means "how do I set up my environment" has no path from the site:

```
tooling/overview/setup/overview.mdx
tooling/overview/setup/mac.mdx
tooling/overview/setup/windows.mdx
tooling/overview/setup/wallet.mdx
tooling/overview/setup/replit.mdx
tooling/overview/setup/development-chain.mdx
tooling/overview/migrate/from-ethereum.mdx
```

The other 15:

```
build-on-celo/support.mdx
build-on-celo/build-on-socialconnect.mdx        (only a redirect *source* mentions it, docs.json:3143)
build-on-celo/build-with-usat.mdx               (no mention anywhere in docs.json)
tooling/contract-verification/foundry.mdx
tooling/oracles/wit-oracle.mdx
tooling/testnets/celo-sepolia/disclaimer.mdx
tooling/wallets/coinbase-wallet.mdx
tooling/wallets/ledger/eip712-workaround.mdx
tooling/libraries-sdks/contractkit/data-encryption-key.mdx
tooling/libraries-sdks/contractkit/migrating-to-contractkit-v1.mdx
tooling/libraries-sdks/contractkit/migrating-to-contractkit-v2.mdx
tooling/libraries-sdks/contractkit/migrating-to-viem.mdx
tooling/libraries-sdks/contractkit/notes-web3-with-contractkit.mdx
legacy/protocol/identity/odis.mdx
legacy/protocol/pos/becoming-a-validator.mdx
```

### Process

A sheet with one row per orphan: path, title, last commit date, audience, proposal (re-nav in its current tab / delete / fold into page X). The team marks decisions asynchronously; one PR applies them. Scope is the orphaned pages only — not an audit of in-nav pages.

### Definition of done

- [x] Sheet shared: https://docs.google.com/spreadsheets/d/1uDxehJlfGSbD1wfcHeROknkIhRdYQZcF_OkY2Sne4LI/edit
[...]

PR #2279, opened 2026-08-24
docs: delete _deprecated/, prune dead-end redirects, add orphan-page check to CI

### LINKED ISSUE #2252: task: Cleanup — delete _deprecated/, prune dead-end redirects, add an orphan-page check to CI
Part of #2266.

### What needs to be done?

Three mechanical cleanups, all verified at `bdf40b37`:

1. **`_deprecated/` holds 105 files with zero references from `docs.json`.** `find _deprecated -type f | wc -l → 105`; `grep -c _deprecated docs.json → 0`. They are copies of `legacy/`, `infra-partners/` and `home/` content made during earlier reorganisations and never deleted. Contributors occasionally edit the copy instead of the live page.
2. **Dead-end redirects.** Of 635 redirects, 223 internal destinations have no matching `.mdx`; 167 of those are also not a redirect source themselves — true dead ends. Examples: `/celo-codebase/protocol/optics → /developer/bridges/bridges`, `/cel2/eclair → /tooling/testnets/`. Also a redirect chain at `docs.json:1468` whose destination `/build/build-with-thirdweb/celo-nft-drop-tutorial` is itself a redirect source (line 3147).
3. **No orphan detection in CI.** Nothing checks that a page on disk is reachable from `navigation`. That is how the docs reached 25 orphaned pages unnoticed (handled in #2253).

### Definition of done

- [ ] `_deprecated/` deleted
- [ ] Dead-end redirects removed or re-pointed to the nearest live page. OPEN: exact count left to implementer — re-derive at fix time; keep any whose destination is a valid non-`.mdx` route (external URL, directory index)
- [ ] Orphan script committed as `scripts/check-orphans.sh`: collect all string leaves of `docs.json` → `navigation`, diff against `find . -name '*.mdx'` excluding `snippets/` and `node_modules/`; prints the list and a count
- [ ] Script added to `docs-validation.yml` as **report-only** (non-failing) — #2253 flips it to failing once the existing orphans are resolved
- [ ] `npx mintlify broken-links` green

### How we'd know it's fixed

`_deprecated/` is gone; the redirect script reports 0 dead ends; every PR's CI log prints the orphan list.

### Non-goals / boundaries

Does not add, delete, or move any orphaned page — #2253 decides them, and the three `build-on-celo/build-with-thirdweb/*` orphans belong to #2255. No content edits.

### Depends on / blocks

No dependencies. Blocks #2254, #2255.

**Measured at:** `bdf40b37`

palango:
`_deprecated/` held 105 unreferenced copies of live pages that contributors occasionally edited by mistake, and 208 of the 645 redirects in docs.json pointed at another redirect or at a path with no page behind it. Nothing in CI noticed when a page on disk fell out of `navigation`, which is how 25 pages went orphaned. This PR deletes `_deprecated/`, repairs the redirect graph so every redirect ends at a live page, and adds the orphan check as a report-only step to the existing docs-validation workflow. Closes #2252.

## Changes

1. `_deprecated/` is deleted. It had zero references from `docs.json` or live content. Before deleting, every file was compared against its closest live counterpart: 30 are byte-identical, the other 75 differ only where the live page moved on (fixed links, updated dates, added descriptions) or where the content is retired L1/Alfajores material. Nothing needed porting. No redirects added, since these paths were never reachable.
2. 53 redirect chains are flattened. AGENTS.md says a redirect points at the final page; these pointed at another redirect's source. Includes the thirdweb chain named in the issue.
3. 152 dead-end redirects are re-pointed and 3 removed. The issue left removal vs re-pointing open; re-pointing keeps old inbound links working, so every redirect whose chain ended at a missing page now points at the live page that answers the same question (most had near-exact matches, e.g. `/cli/account` to `/tooling/libraries-sdks/cli/account`). The 3 removed entries are the `/glossary` group: no glossary page exists anywhere, and they formed a redirect cycle. Every new destination exists on disk, is listed in `navigation`, and is not itself a redirect source.
4. `scripts/check-orphans.sh` diffs `.mdx` files on disk against the page paths under `navigation` in docs.json and exits 1 when it finds orphans. The workflow step has `continue-on-error: true`, so #2253 flips it to failing by deleting that single line.

The counts differ from the issue because the redirects drifted since it was measured at bdf40b37; they were re-derived on the branch as the issue asks, and again after the rebase onto #2271, which re-pointed the MiniPay redirects itself.

## Verification

`bash scripts/check-orphans.sh` prints the 25 known orphans (the count #2253 tracks) and exits 1.

`npx mintlify broken-links`:

```
success no broken links found
```

Re-running the redirect analysis after the rebase: 642 redirects, 0 chains, 0 dead ends. The one remaining wildcard with no pages behind it, `/developer/rainbowkit-celo/:slug*`, predates this PR and is left for #2253.

## Notes for #2253

A few re-pointed redirects have an exact-match target among the 25 orphaned pages (the contractkit migration guides, `tooling/overview/setup/*`), but orphans cannot be redirect targets, so they point at the nearest in-nav page instead. If #2253 adds those pages to `navigation`, the redirects can be tightened.


GigaHierz (dismissed):
Approving after a full verification pass.

- `_deprecated/` deletion is safe: the directory is gone and there are **zero** references to it from `docs.json` or any live `.mdx`.
- Redirect graph is sound: across all 632 redirects there are **0 unflattened chains** (no destination is itself a redirect source) and **0 internal destinations without a backing live page**. Spot-checked the re-pointed classes (e.g. `/cli/* → /tooling/libraries-sdks/cli/*`) and they resolve.
- The new orphan-page CI step is correctly report-only (`continue-on-error: true`); `scripts/check-orphans.sh` runs and honestly lists the existing orphans rather than no-opping.

One pre-existing item unrelated to this PR: the `/developer/rainbowkit-celo/:slug*` wildcard points at a prefix with no pages — untouched here, fine to handle in the orphan/redirect follow-up (#2253).

GigaHierz (approved):
Re-approving after the AGENTS.md cleanup (99e84f6c) dismissed my prior approval. The only change since my full review is dropping the now-deleted `_deprecated/` from the 'Content directories today' list — consistent with this PR deleting the directory. Everything I verified earlier still holds (deletion has zero references, redirect graph has no unflattened chains or missing destinations, orphan CI step is report-only). CI is green.

PR #2291, opened 2026-08-28
docs(thirdweb): collapse seven thirdweb pages into one tool page at parity (#2255)

GigaHierz:
Closes the code half of #2255.

## What changed

thirdweb was documented across **seven pages** — three under Dev Environments (`overview`, `thirdweb`, `one-click-deploy`), an SDK page under Libraries & SDKs, and three orphaned pages under `build-on-celo/build-with-thirdweb/` — with tutorials, step-by-step code, and wording no other tool gets ("Using Thirdweb is recommended because…", "Complete Web3 development framework with powerful wallet tools", "**Recommended**: Use affordable indexers like thirdweb Insight").

It is now **one page**: `tooling/dev-environments/thirdweb`, flat alongside `foundry.mdx`, `hardhat.mdx` and `remix.mdx`, same title pattern ("Deploy on Celo with thirdweb" / sidebar "Using thirdweb"). It carries only what these docs actually own — the Celo chain configuration — and points at thirdweb's own docs for the rest.

### Resolving the ticket's OPEN question

> OPEN: home it under dev-environments or under SDKs; pick one and apply the same choice to the redirects

**Dev Environments.** Foundry, Hardhat and Remix are single flat `.mdx` files in that group; a single page in its own folder next to them would not match. All seven old paths redirect to the one page.

### De-promotion pass on the remaining mentions

| Page | Change |
|---|---|
| `tooling/wallets/index.mdx` | Kept as a plain list item at parity with the other 10 providers; `utm_source=celo&utm_medium=documentation&utm_campaign=chain_docs` dropped (no other provider carries campaign parameters) |
| `build-on-celo/build-with-ai/x402.mdx` | "Alternative: thirdweb facilitator" → "Other facilitators", one sentence and one link; the duplicate Playground row dropped from Resources |
| `build-on-celo/scaling-your-app.mdx` | Indexer section no longer recommends a vendor; points at `/tooling/indexers/overview` |
| `tooling/libraries-sdks/web3/index.mdx` | "Please consider using thirdweb or viem" → "Use viem instead" |
| `tooling/nodes/overview.mdx` | Neutral one-line blurb, utm parameters dropped |
| `build-on-celo/quickstart.mdx` | "Complete Web3 development framework with powerful wallet tools" → "Wallet connector from thirdweb" |

### Two unrelated bugs fixed while in these files

[...]

Issue #2259, opened 2026-08-21
story: Build tab — merge "Build on Celo" and Tooling into build/*, with Agents and Mini Apps leading


GigaHierz:
Part of #2266.

### User story

As an app or agent developer, I want one Build tab ordered **Quickstart → Agents → Mini Apps → Network info → Guides → Tools → Reference**, so the primary paths are one click from the tab root instead of the third group of the second tab.

### Context & evidence

At `bdf40b37`: "Build on Celo" (`docs.json:111`) has 31 pages; "Tooling" (`docs.json:188`) has 90 pages — 31% of all pages in a junk-drawer tab with ten groups. The agent content lives at `build-on-celo/build-with-ai/` (10 pages: `overview`, `use-docs-with-ai`, `8004`, `x402`, `mpp`, `celopedia`, `vibe-coding`, `usecases`, `mcp/index`, `mcp/celo-mcp`) plus `build-on-celo/attribution-tags.mdx`. Naming drifts across "Build with AI" / "Build for MiniPay" / "Build on SocialConnect".

### Scope

Move `build-on-celo/*` and the developer subset of `tooling/*` into `build/*` per the #2209 Appendix A map at `a14395f8926eb43424e0d20e6d67ab96f2c8209b`, with these corrections:

- **Agents** (`build-with-ai/*` incl. `mpp`, `celopedia`, `mcp/*`, plus `attribution-tags`) and **Mini Apps** promoted to top-level groups in the order above
- The map's `build-with-ai/agent-skills` page does not exist — drop it
- thirdweb is one tool page at parity after #2255 — place it in the tools group, not in Guides
- Mini Apps group holds the single MiniPay overview from #2264
- Wallet integration page from #2256; fee-abstraction guide as differentiated in #2257
- `tooling/nodes/run-a-celo-node` → Operate (#2260 owns the destination; this PR only adds the agreed redirect)
- One preposition convention: "Build with X" for tools/SDKs, "Build for X" for platforms; flatten the "Build with AI" sub-groups (Agent Infrastructure / MCP Servers) into one ordered list
- Redirects: catch-alls `/build-on-celo/:slug* → /build/:slug*` and `/tooling/:slug* → /build/tools/:slug*` plus per-file overrides

### Acceptance criteria

- [ ] Tab renamed "Build"; Tooling tab removed
- [ ] Group order: Quickstart → Agents → Mini Apps → Network info → Guides → Tools → Reference
[...]

Issue #2258, opened 2026-08-21
story: Learn tab — move home/* (plus kept L1 content) to learn/*


GigaHierz:
Part of #2266.

### User story

As someone asking "what is Celo", or holding CELO, I want one tab with protocol concepts, governance, history and end-user guidance — instead of a tab called "Home" that is really protocol reference.

### Context & evidence

At `bdf40b37` the "Home" tab (`docs.json:38`) has 30 pages in groups Introduction / Overview / Protocol / Governance / Transactions / Epoch Rewards / Manage applications. `build-on-celo/network-overview.mdx` and `build-on-celo/cel2-architecture.mdx` are protocol reference sitting in the builder tab.

### Scope

- Rename tab **Home → Learn**; move `home/*` → `learn/*` with the group structure from the #2209 Appendix A map at `a14395f8926eb43424e0d20e6d67ab96f2c8209b`: Introduction / Network architecture / Protocol / Governance / Stablecoins & bridging concepts / Epoch rewards / Identity / Using CELO
- Pull `build-on-celo/{network-overview,cel2-architecture}` in (wrong-tab fix)
- Include whatever #2254 classified as "keep → Learn"
- Include the end-user MetaMask / Ledger setup pages as split in #2256
- Redirects: catch-all `{ "source": "/home/:slug*", "destination": "/learn/:slug*" }` plus per-file overrides for every non-1:1 move
- Titles state the audience per `AGENTS.md` (#2251)

### Acceptance criteria

- [ ] Tab renamed; every Learn page reachable from nav
- [ ] Redirects in place: catch-all + overrides; verified by curling 10 old URLs → 200 at the new path
- [ ] Inbound links from other tabs updated
- [ ] `home/celo.mdx` (the landing page) keeps its path or gets a redirect — #2261 decides its final home; do not rewrite it here
- [ ] `npx mintlify broken-links` + orphan check green

### Non-goals

Homepage copy (#2261); content rewrites beyond titles and first paragraphs. CODEOWNERS: `home/` is owned by the blockchain and primitives teams — request their review.

### Depends on / blocks

Depends on #2251, #2252, #2253, #2254, #2256. Do not delete `home/wallets`, `home/exchanges`, `home/ramps` before #2265 lands. Blocks #2261.

**Measured at:** `bdf40b37`

Issue #2257, opened 2026-08-21
task: Fee abstraction — differentiate the guide, contract-reference and spec pages and make their relationship explicit


GigaHierz:
Part of #2266.

### What needs to be done?

Fee abstraction has three in-nav homes that partly restate each other (verified at `bdf40b37`):

| Page | Job |
|---|---|
| `build-on-celo/fee-abstraction/{overview,using-fee-abstraction,add-fee-currency}.mdx` | **Guide** — how to pay gas in a stablecoin, how to register a fee currency (6 redirects point at the overview) |
| `tooling/contracts/fee-currencies.mdx` (13.8 KB) | **Contract reference** — addresses, registry, adapters, per-network tables |
| `specs/fee-abstraction.mdx` | **Protocol spec** — oracle rates, tx-pool rules, intrinsic gas, node flags |

Plus two orphaned copies (`tooling/overview/migrate/from-ethereum.mdx`, `tooling/overview/setup/development-chain.mdx` — #2253) and mentions across 46 other pages (`grep -ril "fee abstraction|fee-abstraction|fee currencies|feeCurrency" --include=*.mdx` → 51 files).

Decided: **all three pages stay.** Each gets one clear job, says so in its first line, and links to the other two instead of restating them.

### Definition of done

- [ ] Each of the three pages opens with a one-line "This page is the X. For Y see …" and cross-links the other two
- [ ] Content that belongs to another page's job is replaced by a link, not copied: the guide loses its address tables (→ reference); the reference loses its how-to prose (→ guide)
- [ ] Titles / sidebar titles differ clearly (e.g. "Fee abstraction guide" / "Fee currency contracts" / "Fee abstraction spec"; OPEN: exact wording)
- [ ] The two orphaned copies reduced to a link to the guide (after #2253 decides whether they survive)
- [ ] Anchor bug fixed: `build-on-celo/fee-abstraction/add-fee-currency.mdx:151` links to `/build-on-celo/fee-abstraction/overview#adapters-for-non-18-decimal-tokens`, but that heading lives in `using-fee-abstraction.mdx` (`mintlify broken-links` does not validate anchors)
- [ ] A troubleshooting section added to the guide per the code-example policy in `AGENTS.md` (#2251): gas in USDC — adapter address vs token address, 6 vs 18 decimals, the `feeCurrency` parameter per SDK
- [ ] Done in place — no directory moves
- [ ] `npx mintlify broken-links` green

### Non-goals

[...]

PR #2293, opened 2026-08-28
docs: resolve the 20 orphaned pages — 10 re-nav'd, 10 deleted with redirects (#2253)

### LINKED ISSUE #2253: task: Orphaned-pages audit — re-nav or delete the 22 pages on disk that are missing from navigation

GigaHierz:
Closes #2253. **Stacked on #2291** — base is `GigaHierz/2255-thirdweb-one-page`, since the orphan check can only flip to gating once the three thirdweb orphans are also gone. Review that one first; this retargets to `main` automatically when it merges.

## What changed

All 20 non-thirdweb orphans resolved per the decisions in the audit sheet (the `Remove` column plus its `Notes`). Every `.mdx` on disk is now reachable from `docs.json` navigation.

### Re-nav'd — current tab, no path moves (10)

| Page | Nav home |
|---|---|
| `tooling/overview/migrate/from-ethereum` | Tooling › Overview |
| `build-on-celo/build-with-usat` | Build on Celo › Use Cases |
| `build-on-celo/build-on-socialconnect` | Build on Celo › Use Cases |
| `tooling/oracles/wit-oracle` | Tooling › Oracles |
| `tooling/wallets/ledger/eip712-workaround` | Tooling › Wallets › Ledger Wallet |
| `contractkit/{data-encryption-key, migrating-to-viem, migrating-to-contractkit-v1, migrating-to-contractkit-v2, notes-web3-with-contractkit}` | Tooling › Libraries & SDKs › ContractKit |

### Deleted with a redirect (10)

The six `tooling/overview/setup/*` pages, `build-on-celo/support`, `tooling/contract-verification/foundry`, `tooling/testnets/celo-sepolia/disclaimer`, `tooling/wallets/coinbase-wallet`. Each redirect points at the nearest page that answers the same question, and none of them point at another redirect source.

## Two places the sheet and the repo disagreed

Worth a look, because in both cases the sheet's inbound-link count was 0 and the repo says otherwise.

**1. `build-on-celo/build-on-socialconnect` — kept, not deleted.** Marked `Remove=TRUE` with inbound = 0, but it is linked from six in-nav pages: `home/wallets.mdx:10`, `tooling/wallets/index.mdx:11`, `build-on-celo/build-with-self.mdx:50`, `build-on-celo/build-on-minipay/overview.mdx:102`, `contractkit/odis.mdx:25`, `contractkit/migrating-to-contractkit-v1.mdx:45`. Re-nav'd into Use Cases next to `build-with-self`; all six links keep working. Please correct the row.

**2. `tooling/contract-verification/foundry` — deleted, but its content was kept.** Also marked inbound = 0; in fact `tooling/dev-environments/foundry.mdx:112` linked to it, and deleting it outright would have left Foundry as the only deploy path in the docs with no verification instructions while Hardhat, Remix, Blockscout and Celoscan all keep theirs. The `forge verify-contract` commands are now a "Using Foundry" section on the contract-verification overview — a page that did not mention Foundry at all before — and the redirect lands there.

Same treatment for the other two deletes that held unique content: the Celo Sepolia terms and conditions became an `<Accordion>` on the testnet page, and the support channels became a "Get support" section on `build-on-celo/index`. Nothing that only existed on a deleted page was dropped.

## Inbound links repointed, not papered over

`tooling/dev-environments/hardhat.mdx` listed three deleted setup pages as its prerequisites; that block is now a single sentence naming the actual prerequisites. `tooling/dev-environments/foundry.mdx:112` points at the new anchor.

## The CI check now gates

`continue-on-error: true` removed from `.github/workflows/docs-validation.yml`, and the stale "report-only" comment updated in `scripts/check-orphans.sh`. A new page that is not added to navigation now fails CI.

## Incidental fixes on pages this already touches

- `og:description` → `description` on the nine re-nav'd pages that carried it. These pages are entering `llms.txt` for the first time and would otherwise arrive with no description at all (#2239).
- `tooling/oracles/wit-oracle.mdx`: emoji heading removed per AGENTS.md, and the Solidity example fixed — `constructor(IWitnetPriceRouter _router))` had an unbalanced paren and a missing brace, so it could not compile. The sheet's "update to sepolia" note is already satisfied: #2278 removed the hardcoded addresses, and the page describes Mainnet contracts with a mainnet RPC.
- `tooling/contract-verification/index.mdx`: the four `/developer/verify/*` links were stale redirect sources; now current paths.

## Verification

```
No orphan pages found.

$ npx mintlify broken-links
success no broken links found
```

Unblocks #2256 (decides `coinbase-wallet` and `ledger/eip712-workaround`), #2257 (decides `from-ethereum` and `development-chain`), #2258 and #2259.


docs.json
```diff
@@ -567,6 +569,46 @@
     "dark": "/images/CeloDocs_LogoDark.svg"
   },
   "redirects": [
+    {
+      "source": "/tooling/overview/setup/overview",
```
palango, line 657: These ten redirects cover the current paths of the deleted pages, but older wildcards further down the array rewrite legacy URLs onto these same paths, and Mintlify only takes one hop. Six URLs that resolved at the base commit now 404:

- `/developer/verify/foundry` (wildcard at line 2585)
- `/network/alfajores/disclaimer`, `/network/baklava/disclaimer`, `/network/celo-sepolia/disclaimer` (2745/2753/2761)
- `/wallet/coinbase-wallet` (2789)
- `/build/support` (the `/build/:slug*` wildcard at 2557 matches before the specific `/build/support` entry at 3149, which is now dead code)

Each needs a specific entry above the wildcard, pointing at the final destination already chosen here. AGENTS.md section 2: point at the final page, not at another redirect.

tooling/overview/migrate/from-ethereum.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Celo for Ethereum Developers
-og:description: Overview of the similarities and differences between the Celo and Ethereum blockchains.
+description: Overview of the similarities and differences between the Celo and Ethereum blockchains
```
palango, line 3: Adding this page to navigation publishes content that contradicts the canonical pages:

- Line 78 says Celo "uses EigenLayer as the data settlement layer". It is EigenDA, for data availability (AGENTS.md section 7, `/operate/specification/eigenda`).
- Line 16 says "1 second block finality" and line 78 "all blocks are finalized immediately... no need to wait for more than 1 block confirmation". AGENTS.md rules out the 1-second-finality phrasing; L2 finality follows Ethereum (`/operate/specification/finality`).
- Line 84 has a reversed link, `(transaction types)[/what-is-celo/...]`, which renders as plain text, and even fixed the path dead-ends: the `/what-is-celo/using-celo/protocol/:slug*` wildcard rewrites to singular `/home/protocol/transaction/...` while the page lives under `transactions/`. The working target is `/home/protocol/transactions/transaction-types`.
- Lines 67-107 still recommend Celo Rosetta and celo-blockchain and describe L1 epoch-block receipt behavior.

Worth a content pass before this enters the sidebar, or hold the page back for #2257.

palango (commented):
Reviewed the full change set against the PR head (6579bf5f). The mechanics mostly check out: all ten new nav entries exist on disk with no duplicates, all ten redirect destinations exist and are in nav, docs.json parses, `check-orphans.sh` passes at head, and the `/build#celo-l2-mainnet` anchor resolves. Fifteen findings survived verification; twelve are inline comments on the diff, and the three below anchor on lines outside it.

Ranked, most severe first: 1. legacy-URL redirect chains (inline, docs.json:573) 2. from-ethereum content errors (inline) 3. migrating-to-viem Alfajores content and broken examples (inline) 4. hardhat prerequisites dead-end (inline) 5. mis-aimed deletion redirects (inline, docs.json:609) 6. socialconnect redirects not tightened (below) 7. orphan-gate blind spots (inline, check-orphans.sh) 8. circular foundry.toml promise (inline) 9. Foundry invisible in list and sidebar (inline) 10. segment-dropping wildcards (below) 11. nonexistent "Edit this page" control (inline) 12. stale duplicate lede (inline) 13. emoji headings on the socialconnect page (inline) 14. wit-oracle fence tags (inline) 15. orphaned image assets (below).

Socialconnect redirects not tightened (docs.json 3085, 2229, 753/761/789/3321). This PR adds `build-on-celo/build-on-socialconnect` to navigation, which was the recorded precondition for re-pointing these. `/build/build-on-socialconnect` (line 3085) is the page's own legacy path and still lands on `/build-on-celo/index`. `/protocol/socialconnect` (2229) still leaves the site for self.xyz, a different product per the Self/SocialConnect split in AGENTS.md section 7. The identity-concept sources at 753, 761 and 3321 still point at the ContractKit ODIS API page, as does 789, whose `#using-the-mapping-for-payment` fragment matches no heading there while the socialconnect page opens with exactly that payments use case. The ODIS-specific sources (769, 821, 3337) are fine as they are.

Segment-dropping wildcards (docs.json 2609, 2613). `/developer/setup/:slug*` and `/developer/migrate/:slug*` both rewrite to `/tooling/overview/:slug*`, dropping a path segment, so all six legacy `/developer/setup/*` URLs and `/developer/migrate/from-ethereum` 404. This predates the PR, but the PR retires the setup pages and publishes from-ethereum, and it forecloses the one-line wildcard fix because the corrected targets are now redirect sources themselves. While the redirect set is open: explicit `/developer/setup/*` entries mirroring the new block at 573-596, plus `/developer/migrate/from-ethereum` and the three celo-for-eth-devs sources (1433, 1513, 1557) pointed at `/tooling/overview/migrate/from-ethereum`, which now answers exactly that question.

Orphaned image assets. Deleting replit.mdx and wallet.mdx leaves `img/doc-images/deploy-replit/` (12 PNGs, 6.9 MB) and `img/doc-images/testnet-wallet/` (6 PNGs, 1.0 MB) with zero references anywhere in the tree. Nothing in CI catches unreferenced images, so `git rm -r` both directories in this PR while the deletions are in context.

tooling/libraries-sdks/contractkit/migrating-to-viem.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Migration document from Contractkit
-og:description: How to migrate from ContractKit to viem.
+description: How to migrate from ContractKit to viem
```
palango, line 4: This page enters the nav with content AGENTS.md rules out and code that does not run:

- Alfajores references at lines 21, 194 and 234 ("Alfajores (44787) is retired - do not reference it"). Line 234's Alfajores USDm address is passed as `feeCurrency` on a mainnet client at line 247, which reverts on chain.
- `tokenAddresses["StableToken]` at lines 156 and 174 is an unterminated string literal.
- `publicClient.getAddresses()` (42), `walletClient.sendSignedTransaction(...)` (80) and `sendTransaction` straight after `simulateContract` (185) are not viem APIs; AGENTS.md: every example that stays must run.
- The title is "Migration document from Contractkit" with no `sidebarTitle`, so that misspelled label is what the sidebar shows, while the sunset banners on the sibling pages call this "the migration guide". `Migrating to viem` would match.

Either fix the page in this PR or hold it out of nav until it is fixed; publishing as-is ships broken examples on the recommended migration path.

tooling/dev-environments/hardhat.mdx
```diff
@@ -14,11 +14,7 @@ How to deploy a smart contract to Celo testnet, Mainnet, or a local network usin
 
 ## Prerequisites
 
-To deploy on Celo using Hardhat, you should have Celo set up Celo in your local environment. If you prefer to deploy without a local environment, you can deploy using Remix or Replit.
-
-- [Using Windows](/tooling/overview/setup/windows)
-- [Using Mac](/tooling/overview/setup/mac)
-- [Using Replit](/tooling/overview/setup/replit)
+Node.js and a package manager installed locally, and an account funded from the [Celo Sepolia faucet](https://faucet.celo.org/celo-sepolia). To deploy without a local environment, use [Remix](/tooling/dev-environments/remix) instead.
```
palango: This sentence replaces the only route readers had to account-creation guidance, and the guide now dead-ends: the config below requires `process.env.MNEMONIC` with derivation path `m/44'/52752'/0'/0`, and after this PR no page in the repo explains how to obtain one. The natural reading of the new prerequisite (MetaMask plus faucet) does not produce a usable mnemonic either; tooling/wallets/metamask/use.mdx says MetaMask cannot export or import along the Celo derivation path. Cheapest fix: link `celocli account:new` (tooling/libraries-sdks/cli/account.mdx, in nav) with a note on the derivation path, or switch the example to a private-key account.

docs.json
```diff
@@ -567,6 +569,46 @@
...
+    {
+      "source": "/tooling/contract-verification/foundry",
+      "destination": "/tooling/contract-verification/index"
+    },
+    {
+      "source": "/tooling/testnets/celo-sepolia/disclaimer",
+      "destination": "/tooling/testnets/celo-sepolia/index"
+    },
+    {
+      "source": "/tooling/wallets/coinbase-wallet",
```
palango, line 693: Three of the new destinations do not answer the question the deleted page answered (AGENTS.md section 2: redirect to the nearest page that answers the same question):

- This one: tooling/wallets/index.mdx never mentions Coinbase and has no chain params. The deleted page's `wallet_addEthereumChain`/`wallet_watchAsset` content survives near-verbatim at `/tooling/wallets/metamask/setup` (in nav), which is the better target.
- `/tooling/overview/setup/wallet` (585) goes to the Celo Sepolia network page, which has faucet links but nothing on creating or importing a wallet; `/tooling/wallets/metamask/add-celo-testnet-to-metamask` covers both.
- `/tooling/overview/setup/replit` (589) goes to a dev-environments index that never mentions Replit or any no-install path; `/tooling/dev-environments/remix` is the surviving browser-only option, as the new hardhat.mdx prerequisite itself says.

scripts/check-orphans.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
-# Report .mdx pages that are not referenced anywhere in the docs.json
-# navigation tree. Report-only in CI for now (issue #2252); issue #2253
-# makes orphans fail CI once the existing ones are resolved.
+# Fail if any .mdx page is not referenced in the docs.json navigation tree.
+# A page on disk that is not in navigation is unreachable from the site.
```
palango, line 3: Now that this fails CI, two coverage gaps in the script become load-bearing:

- The jq expression only collects strings from `pages` arrays. A page referenced via `href` (docs.json line 519 already uses one internally) or Mintlify's `root` key is invisible to it, so the first PR that uses either form gets a false orphan and a red gate whose message misdirects ("add it to navigation" - it is in navigation). Today's run passes only because the one internal href is also duplicated in a pages array. Worth unioning in `href`/`root` string values under `.navigation`.
- `find` only matches `*.mdx`; Mintlify renders `.md` too, so an orphaned `.md` page passes silently. Widening needs paired excludes for README/AGENTS/CLAUDE, so either do that deliberately or note the .mdx-only assumption in this header comment.

tooling/contract-verification/index.mdx
```diff
@@ -1,20 +1,40 @@
...
+forge verify-contract --chain-id 42220 <contract_address> <contract_location> --etherscan-api-key $ETHERSCAN_API_KEY --watch
+```
+
+Celo Sepolia testnet (chain ID 11142220):
+
+```bash
+forge verify-contract --chain-id 11142220 <contract_address> <contract_location> --etherscan-api-key $ETHERSCAN_API_KEY --watch
+```
+
+Passing `--verify` to `forge create` or `forge script` verifies right after deployment, which needs the same Etherscan configuration in `foundry.toml`.
```
palango: This sentence promises "the same Etherscan configuration in foundry.toml", but nothing above it is a foundry.toml configuration, and no `[etherscan]` block exists anywhere in the docs. It matters because tooling/dev-environments/foundry.mdx:112 (also changed in this PR) deep-links to this section specifically for that configuration, so the reader loops between the two claims. It is also not quite right technically: `forge verify-contract` with `--etherscan-api-key`/`ETHERSCAN_API_KEY` needs no foundry.toml at all, and `forge create --verify` resolves known chains (42220 included) from the chain ID plus the env var; an `[etherscan]` block is only needed for a custom verifier URL, which plausibly applies to Celo Sepolia. Either drop the foundry.toml clause here and in foundry.mdx:112, or add a real `[etherscan]` snippet.

tooling/contract-verification/index.mdx
```diff
@@ -1,20 +1,40 @@
...
+Verifying publishes your contract's source so anyone can read it on the block explorer. Pick the route that matches how you deployed.
 
 ## Verify Contracts on Celo
 
-- [Using Blockscout](/developer/verify/blockscout)
-- [Using Remix](/developer/verify/remix)
-- [Using CeloScan](/developer/verify/celoscan)
-- [Using Hardhat](/developer/verify/hardhat)
\ No newline at end of file
+- [Using Hardhat](/tooling/contract-verification/hardhat)
```
palango, line 12: Foundry is missing from this list even though the "Using Foundry" section is two lines below it, and it is also absent from the Contract Verification sidebar group in docs.json, so someone scanning either surface sees four methods and concludes Foundry is not covered. Minimum fix: add `- [Using Foundry](#using-foundry)` here. The symmetric alternative is to keep foundry.mdx as a fifth page in the group alongside Hardhat, Blockscout, Remix and Celoscan.

build-on-celo/index.mdx
```diff
@@ -67,6 +67,15 @@ Using the final audited release, the Celo validator community performed a dry ru
...
 Following a successful Baklava upgrade, the Celo L2 Mainnet officially went live.
 
+## Get support
+
+- General questions: join [Discord](https://discord.com/invite/celo), claim the developer role, and ask in `#build-with-celo`.
+- Advanced questions: open an issue or join [GitHub Discussions](https://github.com/orgs/celo-org/discussions).
+- Longer-form discussion: the [Celo Forum](https://forum.celo.org/).
+- Updates: the [DevDesk mailing list](https://embeds.beehiiv.com/eeadfef4-2f0c-45ce-801c-b920827d5cd2) and [CeloDevs on X](https://x.com/CeloDevs).
+
+Something missing or unclear in these docs? [Open an issue](https://github.com/celo-org/docs/issues/new) or use "Edit this page" at the bottom of any page.
```
palango: docs.json has no `feedback` key, and Mintlify only renders the suggest-edits control when `feedback.suggestEdit` is configured, so there is no "Edit this page" at the bottom of any page; a reader following this instruction finds nothing. The sentence came over from the deleted support page, but since it is being rewritten anyway: either add the `feedback` block to docs.json in this PR or drop the clause and keep the issue link.

tooling/contract-verification/index.mdx
```diff
@@ -1,20 +1,40 @@
 ---
 title: "Verify Contract Deployed on Celo"
 sidebarTitle: "Overview"
-description: How to verify contract deployed on Celo.
+description: How to verify a contract deployed on Celo with Hardhat, Foundry, Remix, Blockscout or Celoscan
 ---
 
-
-
 How to verify contracts deployed on Celo.
```
palango: Leftover from the rewrite: this line plus the `---` under it restate the frontmatter description right above the new lede at line 11, so the page opens with the same sentence twice and a stray rule. Suggest deleting lines 7-9. While in the file, `## Verify Contracts on Celo` on line 13 should be sentence case; the AGENTS.md heading rule explicitly covers rewritten pages.

build-on-celo/build-on-socialconnect.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Introduction to SocialConnect
-og:description: A beginner's guide to understanding and using SocialConnect
+description: How SocialConnect maps phone numbers and other identifiers to Celo addresses
```
palango, line 3: Publishing this page while editing its frontmatter leaves all four headings against the AGENTS.md heading rules this same PR applies to wit-oracle.mdx: emoji in every heading, a question heading on a non-FAQ page, and "Further Reading", which AGENTS.md names as an out-of-vocabulary synonym. The emoji also end up in the generated anchor slugs, so plain `#key-features`-style deep links scroll nowhere, and mint broken-links does not validate anchors. Separately, the page has zero internal links (everything points at socialconnect-docs.vercel.app), which makes the AGENTS.md-canonical page for phone-number mapping a dead end in the link graph.

tooling/oracles/wit-oracle.mdx
```diff
@@ -45,7 +47,7 @@ contract MyCeloContract {
      * the network you are using! Please find available addresses here:
      * https://docs.witnet.io/smart-contracts/witnet-data-feeds/addresses
      */
-    constructor(IWitnetPriceRouter _router))
+    constructor(IWitnetPriceRouter _router) {
```
palango, line 50: Good catch on the constructor. The fence this block sits in is still tagged `javascript` though, as are the other two on the page: lines 36 and 106 are Solidity, and the snippet at line 76 mixes Python calls with `//` comments and parses in neither language. Since this PR already edits inside the first block, worth retagging the fences and fixing the line-76 snippet in the same pass.

GigaHierz:
All 15 addressed in `a3ce702c`. Rebased on main, so #2291 is out of the diff and the base is `main` now.

**Finding 1 + 10 — redirect precedence.** Confirmed the mechanism and it was worse than the six URLs you listed. `/build/:slug*` also shadowed the three thirdweb entries, which is a regression I shipped in #2291: `/build/build-with-thirdweb/overview` hit the wildcard, landed on `/build-on-celo/build-with-thirdweb/overview`, and my redirect from there was a second hop. Four pre-existing `/build/*` sources (composer-kit, mcp, mcp/celo-mcp, mcp/composer-mcp) were dead the same way.

I wrote a resolver that walks the array in order, expands `:slug*`, and checks the destination against the nav set:

```
chains (destination is itself a source):   0
dead destinations (not in navigation):     0
sources shadowed by an earlier wildcard:   0
```

**Finding 2 — from-ethereum.** Every item was right. EigenLayer→EigenDA, the finality phrasing, the reversed link and its non-resolving path. On the two you flagged as L1 leftovers I checked before rewriting: `celo-org/rosetta` is **archived**, and `celo-org/celo-blockchain`'s own description now reads "Previous official repository... for current see op-geth". The precompile link points at `core/vm/celo_contracts.go` in op-geth (`celo-rebase-17`, verified 200) — Celo does still add precompiles, so the claim survives, just not the v1.3.2 link.

**Finding 3 — migrating-to-viem.** All four code defects confirmed. The one worth recording: `0x874069Fa1Eb16D44d622F2e0Ca25eeA172369bC1` passed as `feeCurrency` on a mainnet client —

```
$ eth_call symbol() @ 0x874069... on forno.celo.org  -> no code
$ eth_call symbol() @ 0x765DE8... on forno.celo.org  -> "USDm"
```

so it would revert exactly as you said. Also: the `CeloTokens` helper returned `[string, string][]`, so `tokenAddresses["StableToken"]` was never going to work even with the quote closed. It now returns a `Record` and both blocks use it, which keeps the registry-lookup pattern rather than hardcoding.

**Findings 4, 5, 6, 8, 9, 11, 12, 13, 14** — all applied as suggested. For 5 I took your three targets. For 9 I did the minimum fix (Foundry in the list); keeping `foundry.mdx` as a fifth page is the better shape long-term but that is a nav change for the Build-tab ticket.

**Finding 7 — orphan gate.** Both gaps closed, and I tested them rather than assuming:

```
# orphaned .md is caught
$ touch build-on-celo/zz-temp-orphan.md && bash scripts/check-orphans.sh
Found 1 orphan page(s) ... build-on-celo/zz-temp-orphan     (exit 1)

# href-only reference is not a false positive
$ # removed use-docs-with-ai from its pages array, leaving only the href at :519
No orphan pages found.                                       (exit 0)
```

The `.md` widening needed the excludes you predicted (`README`, `AGENTS`, `CLAUDE`, `scripts/`, dotdirs).

**Finding 15** — both directories gone, 18 files, 7.9 MB.

One thing I did not do: add the `feedback` block to `docs.json` for finding 11. I dropped the clause instead, since turning on suggest-edits site-wide is a bigger call than this PR should make. Worth its own issue if you want the control.

`check-orphans.sh` green, `mintlify broken-links` green.

GigaHierz:
Reopening finding 11 — I got the call wrong, and the answer is more annoying than either of us assumed.

The docs used to have an edit-a-page control and it is wanted back, so "drop the clause" was not the right resolution. But your diagnosis was correct and the fix you offered is not available: `feedback.suggestEdit` was a **mint.json v1** key. Against `https://mintlify.com/docs.json`, which this repo already declares as its `$schema`:

```
$ grep -oiE "feedback|suggestEdit|raiseIssue|thumbsRating" mint-schema.json | sort -u
(no output)
```

Zero hits for all four. There is no `edit` key of any kind in the schema, nothing in Mintlify's own llms.txt about the control, and the live site renders none.

The closest supported mechanism is a custom entry in `contextual.options`, which is now added. One real limitation, which the description is worded around rather than hiding: **the href cannot be per-page.** Mintlify substitutes `$path` only inside query-parameter values, not in the base URL, so `github.com/celo-org/docs/edit/main/<path>.mdx` is not expressible. The entry opens the repository and GitHub's own edit-and-PR flow takes over — which is the workflow that was actually being used.

The GitHub-search alternative (`base: github.com/search`, `query: [{key: q, value: $path}]`) would give a per-page deep link, but GitHub code search requires a login, so a logged-out reader hits a wall instead of a file. Not worth it.

So this PR now carries the site-wide contextual entry after all, which is a bigger footprint than an orphan-audit PR would normally take. Say the word and I will split it out into its own PR — it is one object in `docs.json` plus one sentence in `build-on-celo/index`.

`c5c313a9`. Branch is up to date with main; broken-links and check-orphans both green.

GigaHierz:
Split done — the context-menu change is now **#2297**, and this PR is back to the orphan audit plus its review fixes only.

What stayed here is one sentence on `build-on-celo/index`: the support section no longer names an "Edit this page" control, and instead points at [celo-org/docs](https://github.com/celo-org/docs) and the flow GitHub provides from there. That reads correctly whether or not #2297 lands, so there is no forward reference and the two PRs can merge in either order.

`46500e91`.

tooling/oracles/wit-oracle.mdx
palango: Fences and the constructor are fixed. The web3 snippet parses as Python now, but it still won't run: the ABI string types `_id` as `bytes32` while the call passes four bytes.

```
web3 7.16.0
  bytes32 + bytes.fromhex("9ed884be") -> MismatchedABI
  bytes4  + bytes.fromhex("9ed884be") -> OK  0xd2476bd49ed884be00...
```

The `bytes32` is the wrong half, not the argument. The Solidity block above already calls `router.valueFor(bytes4(0x24beead4))`, and Witnet types these IDs as bytes4 upstream: `IWitPriceFeedsLegacy.sol` declares `latestPrice(bytes4 feedId)` and documents `hash()` as returning "the ERC-2362 identifier of the given data feed caption string, truncated to bytes4". So `"type": "bytes4"` and `"internalType": "bytes4"`, argument unchanged. That moves the selector from `0xf78eea83` to `0xd2476bd4`, so it is not cosmetic; as written the snippet calls a function the proxy does not have.

Same block, smaller: the heading above it still reads `#### Javascript example` over a `python` fence.

One thing beyond this PR. `IWitnetPriceRouter` no longer exists in witnet-solidity-bridge: the interfaces directory now carries `IWitPriceFeeds.sol` with `computeID4`/`getPrice(ID4)`, and the router shape has moved under `legacy/`. Both Solidity blocks on the page import a path upstream dropped. Worth its own issue rather than growing this one.

tooling/overview/migrate/from-ethereum.mdx
```diff
@@ -13,7 +13,7 @@ For a general overview of the Celo network and architecture, see [the Celo Overv
 
 ## What is Celo's Relationship to Ethereum?
```
palango: Required change. Putting this page in the sidebar brings its headings under the AGENTS.md heading rules, the same ones this PR applied to `socialconnect` and `wit-oracle`.

Line 14 is a question heading on a page that is not an FAQ (AGENTS.md:72). Title case is still on lines 14, 23, 58, 89, 93, 99 and 103. AGENTS.md:71 does say not to mass-retitle old pages for case alone, but it scopes that to pages you are not otherwise touching, and this PR rewrites four sections of this one.

docs.json
```diff
@@ -2178,7 +2312,7 @@
     },
     {
       "source": "/protocol/socialconnect",
```
palango, line 2317: Required change. `/protocol/identity/index` at line 2230 still points at `https://self.xyz/`. With this line fixed the two legacy identity paths now disagree: `/celo-codebase/protocol/identity/index` reaches the SocialConnect page this PR adds to nav, while `/protocol/identity/index` leaves the site for a different product. AGENTS.md section 7 keeps Self and SocialConnect apart, so self.xyz does not answer the identifier-mapping question. Give it the same destination as this line.

palango (approved):

Issue #2298, opened 2026-08-31
task: Audit historical architecture changes and add missing redirects (Page Not Found is the #4 most-viewed page in GA4)


GigaHierz:
## What needs to be done

Add redirects for old public URLs that still receive traffic but now **404**, and establish redirect coverage for every past architecture change (Docusaurus→Mintlify, `infra-partners/`→`operate/`, Legacy/`_deprecated` removal, MiniPay collapse, Specs fold-in, etc.).

**Evidence (GA4 — `docs.celo.org` property, Aug 3–30):** the **"Page Not Found"** page is the **#4 most-viewed page** — 74 views (4.4% of all pageviews), 33 active users (6.4%). GA4's page-*title* view can't reveal *which* paths failed, so coverage was reconstructed from git history instead.

**Audit (full `git log --all`, `origin/main` @ `2ed9dba`):**
- 2,273 historical public URLs reconstructed across both eras (confirmed Docusaurus `routeBasePath: "/"` → there is **no** `/docs/` URL prefix).
- 237 covered by a live page; 770 covered by an existing redirect.
- **1,266 currently 404 with no redirect**, broken down:
  - **449** — generated GitBook/typedoc API dumps under `/developer-resources/**/reference/**`
  - **415** — old Docusaurus `/blog/**`
  - **107** — old `/v2/**` GitBook tree
  - **295** — individual content pages (~80 with a confident target, ~215 needing an editorial target)
- Live spot-check confirmed real 404s (e.g. `/about-celo/gas-fees`, `/protocol/identity/odis`, `/developer/dev-environments/hardhat`); live controls returned 200.

## Scope / slices

1. **33 high-confidence redirects** — every destination verified to be a live page on `main`; led by `/about-celo/:slug* → /home/:slug*` (covers 26 pages). **A PR is open for this slice.**
2. **3 bulk catch-all wildcards — target decision needed (OPEN):**
   - `/developer-resources/:slug*` → ?  (retires 449 dead API-dump URLs)
   - `/blog/:slug*` → ?  (415 removed blog posts)
   - `/v2/:slug*` → ?  (107 old GitBook pages)

   Suggested defaults: `blog` + `v2` → `/` (content retired); `developer-resources` → `/tooling/libraries-sdks/contractkit`. **OPEN (implementer/PM to confirm): pick targets, or intentionally 410/leave.**
[...]

PR #2299, opened 2026-08-31
docs(redirects): add 51 verified redirects for pre-migration URLs still 404ing

GigaHierz:
## Why

GA4 shows the **"Page Not Found"** page is the **#4 most-viewed page** on docs.celo.org (~4.4% of pageviews). An audit of the full git history (`git log --all`, both the Docusaurus and Mintlify eras) found **~1,266 old public URLs that 404 with no redirect**. See #2298 for the full breakdown.

## What this PR does

Adds **51 redirect sources**, `docs.json` only, `737 → 788` entries. Every destination is a live page on `main`.

- `/about-celo/:slug* → /home/:slug*` — one wildcard covering the `about-celo/`→`home/` rename, with `/about-celo`, `/about-celo/manage/exchange` and `/about-celo/protocol/consensus` placed **above** it because the wildcard alone sends those three to 404s.
- `/developer/dev-environments/{hardhat,foundry,remix,thirdweb}` → `/tooling/dev-environments/*` (not caught by existing `/developer/*` wildcards).
- The nine `/protocol/identity/*` siblings, mirroring the `/legacy/protocol/identity/*` destinations already on `main`.
- The four testnet landing pages, `/oracle`, `/network/cannoli`, and old `holder/manage/`, `community/`, `general/` paths.

## What it does NOT do (tracked in #2298)

- **The 3 bulk buckets** — old `/blog/**` (415), `/v2/**` (107), generated `/developer-resources/**` API dumps (449). An earlier revision of this branch added these as catch-all wildcards; that commit has been **dropped**, so the PR now matches this section. The review found two concrete reasons the wildcards were wrong as written, both recorded on #2298 so the analysis is not lost.
- ~215 individual paths with no confident target — need editorial triage.

## Verification

```
$ python3 -c "import json;print(len(json.load(open('docs.json'))['redirects']))"
788                                    # main: 737

$ mint broken-links
success no broken links found
```

Checked programmatically over the whole array on this head:

- **0 duplicate sources.**
- **0 chains** — no destination is itself a redirect source. (Three were, before the review; see below.)
- **0 dead destinations** — every internal destination resolves to a real `.mdx` on disk.
- Diff is redirect entries only; no formatting churn elsewhere in `docs.json`.

Live status of all 51 sources against production:

| Result | Count |
|---|---|
| 404 today (the ones this fixes) | 48 |
| Already resolve (200) | 3 |

The three that already resolve are `/community/release-process`, `/developer-resources/contractkit` and `/developer-resources/forno`. Something outside `docs.json` handles them; explicit entries are still preferable, but this confirms the audit's source list was not live-checked. Full sweep replaces the earlier spot-check.

Refs #2298


docs.json
```diff
@@ -3415,6 +3415,150 @@
...
+    {
+      "source": "/cel2/fee-currencies",
+      "destination": "/tooling/contracts/fee-currencies"
+    },
+    {
+      "source": "/build-on-celo/build-with-ai/celina",
+      "destination": "/build-on-celo/build-with-ai/mcp/celina"
+    },
+    {
+      "source": "/network/cannoli/disclaimer",
```
palango, line 3668: All four of these disclaimer redirects land on a redirect rather than a page, so all four 404. `/tooling/testnets/celo-sepolia/disclaimer` was deleted by #2293 (merged as `bb57fda6`) and is now itself a source, pointing at `/tooling/testnets/celo-sepolia/index`. Mintlify takes one hop. AGENTS.md section 2: point at the final page, not at another redirect. Give all four `/tooling/testnets/celo-sepolia/index`, the destination #2293 already chose for the `/network/*/disclaimer` sources.

docs.json
```diff
@@ -3415,6 +3415,150 @@
...
+    {
+      "source": "/general/history",
+      "destination": "/home/history"
+    },
+    {
+      "source": "/general/using-celo/gas-fees",
+      "destination": "/home/gas-fees"
+    },
+    {
+      "source": "/general/ecosystem/daos",
```
palango, line 3564: Same chain. `/contribute-to-celo/daos` was deleted by #2301 (`6136d331`) and now redirects to `/contribute-to-celo`. Every sibling already skips the hop: `/general/ecosystem/builders`, `/general/ecosystem/contributors`, `/general/ecosystem/overview` and `/what-is-celo/joining-celo/daos` all point at `/contribute-to-celo` on main. This one should too.

docs.json
```diff
@@ -3415,6 +3415,150 @@
...
+    {
+      "source": "/tooling/testnets/mainnet/disclaimer",
+      "destination": "/tooling/testnets/celo-sepolia/disclaimer"
+    },
+    {
+      "source": "/blog/:slug*",
+      "destination": "/"
+    },
+    {
+      "source": "/v2/:slug*",
```
palango: This throws away mappings the first commit writes by hand. `docs/v2/about/our-history.md`, `docs/v2/about/using-celo/gas-fees.md` and `docs/v2/about/ecosystem/daos.md` are the same three pages that commit maps as `/general/history`, `/general/using-celo/gas-fees` and `/general/ecosystem/daos`. Under the wildcard, a reader arriving from `/v2/about/using-celo/gas-fees` gets the homepage while the `/general/` twin gets the gas-fees page. Either add the handful of `/v2/about/*` entries above the wildcard, or hold the wildcard until the bucket is triaged.

docs.json
```diff
@@ -3415,6 +3415,150 @@
...
+      "destination": "/"
+    },
+    {
+      "source": "/developer-resources/:slug*",
```
palango: 501 files lived under `developer-resources/` historically, 141 of them under `developer-resources/contractkit/`. Sending all of them to one landing page is a long way from the nearest page that answers the same question. One entry above the catch-all covers the largest bucket: `/developer-resources/contractkit/:slug*` to `/tooling/libraries-sdks/contractkit`.

palango (commented):
Reviewed against the PR head (f83df743), checking every destination against `origin/main` rather than the branch working tree, since the branch is based on `2ed9dba9` and main is at `6136d331`. The file itself is sound: `docs.json` parses, `mint broken-links` passes, there are no duplicate sources in the array, none of the 36 new sources collides with anything on main, and the three catch-alls are the last three entries with every specific source above the wildcard that would otherwise swallow it. Ten of the twelve sources I sampled do 404 today.

Twelve findings. Nine are inline on the diff; the three below anchor outside it.

Ranked, most severe first: 1. four disclaimer redirects chain (inline, 3536) 2. `/general/ecosystem/daos` chains (inline, 3432) 3. the branch predates both merges that caused 1 and 2 (below) 4. the second commit does what the body says this PR does not (below) 5. `/v2/:slug*` discards mappings the first commit writes by hand (inline, 3556) 6. `/developer-resources/:slug*` swallows 501 files (inline, 3560) 7. `/about-celo/:slug*` misses two of its own children (inline, 3420) 8. four testnet landing URLs still 404 (inline, 3540) 9. `/oracle` and `/network/cannoli` indexes get no entry (inline, 3484) 10. eight `/protocol/identity/*` siblings left 404ing (inline, 3480) 11. `.../usage` resolves two ways (inline, 3468) 12. two sampled sources already return 200 (below).

Branch base (findings 1 and 2 exist only because of it). The branch is based on `2ed9dba9`; main is at `6136d331`. #2293 (`bb57fda6`) and #2301 (`6136d331`) both rewrote `docs.json` in between, and both chained destinations were live at the branch base and are gone on main. GitHub still calls this MERGEABLE, because the new entries append to the end of the array while main's changes sit mid-array, so nothing flags it at merge time. Rebase, then re-check each destination against main.

The second commit does what the body says this PR does not. The body has a section headed "What it does NOT do (tracked in #2298)" naming "The 3 bulk buckets (old `/blog/**` 415, `/v2/**` 107, generated `/developer-resources/**` API dumps 449) — those need a target decision". Commit f83df743 adds all three as catch-alls. The counts disagree too: the body says 33 redirects and 714 → 747, the branch has 36 and 714 → 750, and the difference is those three entries. Either split the commit into its own PR against #2298, or rewrite the body to own the decision.

Two sampled sources already resolve. Of the twelve I checked against the live site, `/community/release-process` and `/developer-resources/contractkit` return 200 today, on `/contribute-to-celo/release-process` and `/tooling/libraries-sdks/contractkit`. Nothing in main's `docs.json` matches either, so something outside the file is handling them and I cannot tell you what. Explicit entries beat whatever that is, but the same method produced the other 31 sources, so check how the audit sampled.

docs.json
```diff
@@ -3415,6 +3415,150 @@
     {
       "source": "/specs/:slug*",
       "destination": "/operate/specification/:slug*"
+    },
+    {
+      "source": "/about-celo/:slug*",
```
palango, line 3552: This wildcard misses two of its own children. I mapped every historical `about-celo/` path onto `/home/<slug>`: 27 of 29 reach a live page that is in navigation. `/about-celo/manage/exchange` and `/about-celo/protocol/consensus` do not, and no earlier entry covers them, so both become a redirect to a 404. Main already records the right targets: `/holder/exchange` points at `/home/manage/asset`, and `/celo-codebase/protocol/consensus` at `/home/protocol/security-council`.

There is also no bare `/about-celo` entry. `about-celo/index.mdx` existed, so the bare URL is real inbound traffic, and a `:slug*` wildcard matches the bare prefix with an empty slug, producing `/home/` with a trailing slash. The two pairs immediately above show the fix twice over: `/infra-partners` beside `/infra-partners/:slug*`, `/specs` beside `/specs/:slug*`.

One ordering note. This wildcard sits mid-array with 32 specific entries after it. Nothing breaks today, but the two entries above have to go in above this line, and appending them at the bottom of the array, where everything else in this PR went, silently does nothing. Move the wildcard to the end of the block.

docs.json
```diff
@@ -3415,6 +3415,150 @@
...
+    {
+      "source": "/build-on-celo/build-with-ai/celina",
+      "destination": "/build-on-celo/build-with-ai/mcp/celina"
+    },
+    {
+      "source": "/network/cannoli/disclaimer",
+      "destination": "/tooling/testnets/celo-sepolia/disclaimer"
+    },
+    {
+      "source": "/tooling/testnets/alfajores/disclaimer",
```
palango, line 3672: The disclaimers get redirects and the landing pages do not. `tooling/testnets/alfajores/index.mdx`, `baklava/index.mdx`, `mainnet/index.mdx` and `eclair/index.mdx` all existed, added in `d09e2b60`, and main carries no `/tooling/testnets` source apart from the celo-sepolia disclaimer. A testnet landing page collects more inbound links than its disclaimer does. Add the four bare paths pointing at `/tooling/testnets/celo-sepolia`, matching the `/network/eclair/:slug*` convention already in the file.

docs.json
```diff
@@ -3415,6 +3415,150 @@
...
+    {
+      "source": "/developer-resources/forno",
+      "destination": "/tooling/nodes/forno"
+    },
+    {
+      "source": "/protocol/identity/odis",
+      "destination": "/tooling/libraries-sdks/contractkit/odis"
+    },
+    {
+      "source": "/oracle/band-protocol",
```
palango, line 3616: `/oracle` gets no entry while all three of its children do. `docs/oracle/index.md` existed alongside `band-protocol.md`, `redstone.md` and `run.md`, and `docs/network/cannoli/index.md` is in the same position relative to the cannoli disclaimer below. A category index is usually the most linked URL in its group, so these are the 404s the audit set out to close. `/oracle` to `/tooling/oracles`, `/network/cannoli` to `/tooling/testnets/celo-sepolia`.

docs.json
```diff
@@ -3415,6 +3415,150 @@
...
+    {
+      "source": "/developer/dappkit/usage",
+      "destination": "/tooling/libraries-sdks/contractkit/usage"
+    },
+    {
+      "source": "/developer-resources/forno",
+      "destination": "/tooling/nodes/forno"
+    },
+    {
+      "source": "/protocol/identity/odis",
```
palango, line 3612: Eight siblings in this directory are left 404ing. History has `docs/protocol/identity/` holding `odis-domain`, `odis-domain-sequential-delay-domain`, `odis-use-case-key-hardening`, `odis-use-case-phone-number-privacy`, `encrypted-cloud-backup`, `metadata`, `privacy-research`, `randomness` and `smart-contract-accounts`. Main already carries the destination for every one of them under the `/legacy/protocol/identity/` and `/celo-codebase/protocol/identity/` prefixes. Copying that block with the `/protocol/` prefix is mechanical.

docs.json
```diff
@@ -3415,6 +3415,150 @@
...
+    {
+      "source": "/holder/manage/self-custody",
+      "destination": "/home/manage/self-custody"
+    },
+    {
+      "source": "/developer-resources/contractkit",
+      "destination": "/tooling/libraries-sdks/contractkit"
+    },
+    {
+      "source": "/developer-resources/dappkit/usage",
```
palango, line 3600: These two send `.../usage` somewhere different from the entries already on main. Main maps `/developer-guide/contractkit/usage` and `/developer-resources/contractkit/usage` to `/tooling/libraries-sdks/contractkit/setup`; these map `/developer-resources/dappkit/usage` and `/developer/dappkit/usage` to `/tooling/libraries-sdks/contractkit/usage`. Both pages exist, so nothing 404s, but two readers following the same old ContractKit usage link from different eras land on different pages. Pick one and make the four agree.

GigaHierz:
Rebased onto `main` (`36622e58`) and worked all twelve. Two commits now: the original 33, plus one addressing the review.

**3 — branch base.** Rebased. The catch-all commit is dropped (see 4), so the branch is `main` + the redirect set.

**4 — the second commit does what the body says this PR does not.** Took the split option: commit `f83df743` is gone from the branch and the body no longer claims 33 (it is 51 sources, `737 → 788`). The wildcards were not merely out of scope — findings 5 and 6 show they were wrong as written — so I recorded both on #2298 rather than carrying them here.

**1 — four disclaimer redirects chain.** Fixed. All four now point at `/tooling/testnets/celo-sepolia/index`, matching what #2293 chose for the `/network/*/disclaimer` sources.

**2 — `/general/ecosystem/daos` chains.** Fixed, now `/contribute-to-celo`, matching its three siblings. This one was my own doing — #2301 created the hop.

**7 — `/about-celo/:slug*` misses two of its own children.** Added `/about-celo/manage/exchange → /home/manage/asset` and `/about-celo/protocol/consensus → /home/protocol/security-council`, plus the bare `/about-celo → /home`. All three are inserted **above** the wildcard in the array, not appended, so they actually take effect.

**8 — four testnet landing URLs.** Added, all four to `/tooling/testnets/celo-sepolia/index`.

**9 — `/oracle` and `/network/cannoli`.** Added, to `/tooling/oracles` and `/tooling/testnets/celo-sepolia/index`.

**10 — eight `/protocol/identity/*` siblings.** Added nine (the eight plus `metadata`), destinations copied from the `/legacy/protocol/identity/*` block already on main.

**11 — `.../usage` resolves two ways.** Aligned the two entries this PR adds to main's convention (`→ /tooling/libraries-sdks/contractkit/setup`) rather than editing main's two, so the change stays inside this PR's own additions. All four now agree. Both `setup.mdx` and `usage.mdx` exist and are in nav, so if you would rather all four point at `usage`, that is a two-line follow-up.

**5, 6 — the `/v2/` and `/developer-resources/` wildcards.** Moot here, recorded on #2298 with your specific remedies.

**12 — sources that already resolve.** You sampled 12 and found 2. I checked all 51 against production:

| | |
|---|---|
| 404 today | 48 |
| Already 200 | 3 |

The third is `/developer-resources/forno`. So ~6% of the audit's sources were already handled by something outside `docs.json`. The audit derived sources from git history without a live check — worth fixing in the method before the remaining ~215 are triaged; noted on #2298.

Worth flagging: my first sweep reported all 51 as 404. That was a bug in my own test script, not the site — BSD `sed` does not support `\?`, so every URL was requested with a stray `",` appended. The table above is from the corrected run.

```
$ mint broken-links
success no broken links found
```

Also checked over the whole array, not just the diff: 0 duplicate sources, 0 chains, 0 destinations without a page on disk. The diff is redirect entries only.

palango (approved):
All twelve findings from the previous round are addressed. Re-checked against head `14c2efee` with destinations resolved on `origin/main` (`25aca481`).

The four disclaimer redirects and `/general/ecosystem/daos` now point at final pages, the branch is rebased past the two merges that caused those chains, and the three catch-all wildcards are gone along with the commit that added them, so the body and the diff agree again. The `/about-celo/:slug*` wildcard has its two exceptions and the bare path above it and now sits at the end of the block. Testnet landings, `/oracle`, `/network/cannoli`, the nine `/protocol/identity/*` siblings and the two dappkit `usage` entries all landed as discussed, with the identity destinations matching main's `/legacy/` twins exactly.

Verified independently: `docs.json` parses, 737 to 788 with 51 new sources, no duplicates, no chains against any specific source or `:slug*` prefix in the array, every non-wildcard destination backed by a real file, no new source shadowed by an earlier wildcard, and all 29 historical `about-celo/` paths reaching a live page. Live sample of ten sources matched the body. CI green.

Two non-blocking notes. The branch is behind main by `e1f87763` and `25aca481`; neither touches `docs.json`, so Update branch is enough before merge. And the PR mixes bare group destinations with `/index`-suffixed ones, which main already does in both directions, so it is pre-existing noise rather than anything this PR introduces.

PR #2300, opened 2026-08-31
feat(analytics): route GA4 through GTM and instrument engagement events

GigaHierz:
## What

Upgrades docs.celo.org measurement beyond bare page views. One concern: analytics instrumentation.

- `docs.json`: swap `integrations.ga4` for `integrations.gtm`. The GA4 Google Tag (same property, `G-0CXEKQ81V2`, so reporting history stays continuous) moves inside the GTM container; keeping both would double-count page views. Container `GTM-NP9GP2BT` is created and published with the GA4 Google Tag.
- `snippets/AddNetworkButton.jsx`: push `dataLayer` events for the add-network conversion — `add_network_click` and `add_network_result` (`success` / `rejected` / `error` / `no_wallet`). Highest-intent action on the site, previously untracked.
- `ANALYTICS.md` (new, internal — not in `navigation`, so not a public page): architecture and three runbooks — GTM container tags/triggers, GA4 property config (AI-assistant channel group, custom dimensions, exit-page explorations), and Cloudflare AI Crawl Control for bot/agent traffic, which no client-side tag can see (AI crawlers don't execute JS).
- `AGENTS.md`: one-line pointer to ANALYTICS.md.

## Why

The site had a single GA4 measurement ID and nothing else: no referrer/drop-off/engagement instrumentation and zero visibility into AI-agent traffic, which reads `llms.txt` and `.md` variants directly. GTM (free) is the only sanctioned route to custom event tracking on Mintlify hosting without a plan upgrade; Cloudflare (free) is the only layer that can count non-JS bots.

## Before merge

- [x] **GTM container has a GA4 destination.** Was wrong when this PR was opened — the container's only Google tag carried the container's own ID as its Tag ID, so merging would have stopped page-view collection silently. Fixed in the GTM UI and re-verified against the published container:

  ```
  $ curl -s "https://www.googletagmanager.com/gtag/js?id=GTM-NP9GP2BT" | grep -c "G-0CXEKQ81V2"
  1

  $ curl -s "https://www.googletagmanager.com/gtm.js?id=GTM-NP9GP2BT"
    "tags":[{"function":"__googtag","vtp_tagId":"G-0CXEKQ81V2","tag_id":4},{"function":"__hl","tag_id":5}]
    "predicates":[ gtm.init, gtm.historyChange, gtm.js ]
    "rules": tag 4 fires on BOTH gtm.init and gtm.historyChange
  ```

  The history-change rule matters on Mintlify: without it only the first page of each visit is counted. It is present.

- [ ] Ownership table in ANALYTICS.md filled in
- [ ] The six runbook-1 event tags created. Until they exist, the `add_network_click` / `add_network_result` `dataLayer` pushes this PR ships fire into a void — none of the six event names appears in the published container today.

## Verification

`mint broken-links` on the branch head:

```
success no broken links found
```

`snippets/AddNetworkButton.jsx` syntax-checked with esbuild (compiles clean). No pages moved; no redirects needed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,7 @@ Structural work in progress is tracked in the restructure epic, [#2266](https://
 - `snippets/` holds reusable JSX/MDX (`/snippets/ColoredText.jsx`, `/snippets/YouTube.jsx`, `/snippets/AddNetworkButton.jsx`). Import with an absolute path after the frontmatter: `import {YouTube} from '/snippets/YouTube.jsx'`.
 - Static assets: `img/`, `images/`, `assets/`, `logo/`.
 - **Any `.js` file under the content root runs on every published page.** Mintlify injects them automatically — there is no allowlist and no way to scope one to a single page — and the same applies to `.css`. Treat a `.js` file here as production code shipped to every reader, not as content: it has full same-origin DOM access on pages that print contract addresses and RPC endpoints. Mintlify does not support a raw `<script src>` in MDX, so third-party scripts are injected programmatically from such a file (`assistant.js` is the example). Note `submodules/developer-tooling` sits under this root too.
+- Site analytics (GA4 via Google Tag Manager for humans, Cloudflare AI Crawl Control for bots/agents) are documented in [ANALYTICS.md](./ANALYTICS.md).
 
 ### Commands
 
--- a/ANALYTICS.md
+++ b/ANALYTICS.md
@@ -21,7 +21,7 @@ So measurement is split in two:
 | Asset | ID | Owner |
 |---|---|---|
 | GA4 property | `G-0CXEKQ81V2` | _fill in_ |
-| GTM container | `GTM-XXXXXXX` (placeholder — replace in `docs.json` when created) | _fill in_ |
+| GTM container | `GTM-NP9GP2BT` | _fill in_ |
 | Cloudflare zone / DNS for `docs.celo.org` | — | _fill in_ |
 | Mintlify org (Free/Starter plan) | — | _fill in_ |
 
--- a/ANALYTICS.md
+++ b/ANALYTICS.md
@@ -26,7 +26,7 @@ already assigns `/docs.json` to the same team. Request access through the team.
 | GA4 property | `G-0CXEKQ81V2` | `@celo-org/devrel` |
 | GTM container | `GTM-NP9GP2BT` | `@celo-org/devrel` |
 | Cloudflare zone / DNS for `docs.celo.org` | `celo.org` zone, `docs.celo.org` proxied | `@celo-org/devrel` |
-| Mintlify org (Free/Starter plan) | — | `@celo-org/devrel` |
+| Mintlify org | Starter plan (established in #2250) | `@celo-org/devrel` |
 
 The permission that matters when a tag needs fixing is **Publish** on the GTM container;
 Edit rights cannot ship a change.
@@ -90,4 +90,6 @@ Cloudflare Analytics (total requests) vs GA4 (sessions) also gives a rough overa
 
 - **Referrer-less AI traffic**: many clicks out of ChatGPT/Claude/Perplexity carry no referrer and appear as Direct in GA4.
 - **JS-capable agentic browsers** (Comet, Atlas, computer-use agents) execute the GA4 tag and count as humans; `is_automated` catches only naive automation.
-- **MCP server queries** (`https://docs.celo.org/mcp`) are served by Mintlify and visible in neither GA4 nor Cloudflare page analytics. Mintlify's own analytics dashboard (human-vs-agent split, assistant/search analytics) requires a paid Mintlify plan — the current plan is Free/Starter; upgrading is the future option if MCP/assistant visibility becomes important.
+- **MCP query content** — not MCP volume. Requests to `https://docs.celo.org/mcp` *do* traverse Cloudflare (they return `cf-ray`, `cf-cache-status: DYNAMIC`), so request volume is visible in Cloudflare's HTTP analytics filtered by path. They are invisible to GA4, which needs JavaScript. What no layer here can show is **what was asked**: the question text, which tool was called, whether the answer cited anything.
+
+  Mintlify's own dashboard would cover part of that, but it needs the Pro plan ($450/mo), which #2250 evaluated and declined — the site was previously on Pro and deliberately moved to Starter. The intended source of query-level signal is instead the in-page assistant (#2286), which can log its own questions. That telemetry does not exist yet, so the docs-gap loop the assistant was chosen for is not closed.
--- a/ANALYTICS.md
+++ b/ANALYTICS.md
@@ -18,12 +18,18 @@ So measurement is split in two:
 
 ## Ownership
 
+Owners are teams, not individuals — this repository is public, and `.github/CODEOWNERS`
+already assigns `/docs.json` to the same team. Request access through the team.
+
 | Asset | ID | Owner |
 |---|---|---|
-| GA4 property | `G-0CXEKQ81V2` | _fill in_ |
-| GTM container | `GTM-NP9GP2BT` | _fill in_ |
-| Cloudflare zone / DNS for `docs.celo.org` | — | _fill in_ |
-| Mintlify org (Free/Starter plan) | — | _fill in_ |
+| GA4 property | `G-0CXEKQ81V2` | `@celo-org/devrel` |
+| GTM container | `GTM-NP9GP2BT` | `@celo-org/devrel` |
+| Cloudflare zone / DNS for `docs.celo.org` | `celo.org` zone, `docs.celo.org` proxied | `@celo-org/devrel` |
+| Mintlify org (Free/Starter plan) | — | `@celo-org/devrel` |
+
+The permission that matters when a tag needs fixing is **Publish** on the GTM container;
+Edit rights cannot ship a change.
 
 ## What is instrumented in this repo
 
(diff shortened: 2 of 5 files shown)


GigaHierz:
The GA4 destination is now wired into `GTM-NP9GP2BT` and the container is republished. Re-verified against the live container, not the GTM UI:

```
$ curl -s "https://www.googletagmanager.com/gtag/js?id=GTM-NP9GP2BT" | grep -c "G-0CXEKQ81V2"
1

$ curl -s "https://www.googletagmanager.com/gtm.js?id=GTM-NP9GP2BT"
  "tags":[{"function":"__googtag","vtp_tagId":"G-0CXEKQ81V2","tag_id":4},{"function":"__hl","tag_id":5}]
  "predicates":[ gtm.init, gtm.historyChange, gtm.js ]
  "rules": tag 4 fires on both gtm.init and gtm.historyChange
```

`vtp_tagId` was `GTM-NP9GP2BT` — the container's own ID — when this PR was opened, so the swap from `integrations.ga4` to `integrations.gtm` would have stopped collection with no error anywhere. It now carries the measurement ID, and the history-change rule is present, which is what keeps SPA navigations counted on Mintlify.

**Page-view parity is safe to merge.** Two things are still outstanding and neither blocks it:

1. **None of the six event tags exists yet** — the container holds 2 tags (the Google tag and the history listener). `add_network_click`, `add_network_result`, `scroll_depth`, `outbound_click`, `copy_code`, `ai_menu_click` and `is_automated` all return 0 occurrences in the published payload. Until they exist the `dataLayer` pushes in `snippets/AddNetworkButton.jsx` fire into a void, so the PR ships page-view parity plus dormant instrumentation rather than the engagement data it describes. Runbook 1 in ANALYTICS.md covers each one.
2. **Ownership table in ANALYTICS.md** still has four `_fill in_` rows.

Also in this PR since the last review: `scripts/check-orphans.sh` now excludes `ANALYTICS.md`. The orphan check began enforcing when #2293 removed its `continue-on-error`, and it was failing CI on a file that is deliberately not in navigation — the same category the script already excludes for `README.md`, `AGENTS.md` and `CLAUDE.md`.

```
No orphan pages found.
$ mint broken-links
success no broken links found
```

palango (changes requested):
The diff itself is fine, and page-view collection keeps working after the swap. What stops me approving is that ANALYTICS.md ticks off four things as confirmed that aren't true, and the swap turns off the assistant's telemetry without the PR mentioning it.

I checked the claims by loading three pages in headless Chrome and capturing the `/g/collect` hits GA4 actually sends: the live site as it is today, a page with only a plain gtag.js snippet for `G-0CXEKQ81V2`, and a page with only the `GTM-NP9GP2BT` snippet. Each hit carries a session sequence number (`_s`), so duplicates below are distinct hits, not retries.

| Setup | `page_view` per load | `page_view` per `pushState` |
|---|---|---|
| Live site today (`integrations.ga4`) | 2 | 2 |
| Plain gtag.js, no GTM | 1 | 1 |
| `GTM-NP9GP2BT` as published | 1 | 2 |

### 1. The History Change trigger double-counts every in-site navigation

GA4 enhanced measurement already sends a `page_view` on `pushState` for this stream. The plain gtag.js page, with no GTM and no history trigger, sends exactly one per navigation, with `dr` set to the previous page. Under the container the Google tag fires a second time on `gtm.historyChange`, so each navigation produces two hits: one from the re-run config about 2 s after the push (no `dr`), one from enhanced measurement about 5 s later (with `dr`).

ANALYTICS.md:47 says the opposite ("without the history trigger only the first page view is counted") and marks it done. The History Change trigger needs to come off the Google tag, and that line needs to change. The live site already double-counts today through Mintlify's own integration, so this isn't a regression, but as written the runbook tells the next maintainer to keep the wrong setup.

### 2. The swap removes `window.gtag`, and the assistant goes quiet

On the live site `typeof window.gtag` is `function`. On a page carrying only the GTM container it is `undefined`. `public/widget.js` in docs-ai-assistant only tracks `if (typeof window.gtag === 'function')`, so all seven `assistant_*` events stop the moment this merges. I also tried the obvious workaround, a `gtag(){ dataLayer.push(arguments) }` shim on the GTM page followed by `gtag('event', 'assistant_opened')`. No hit went out. A shim in the widget won't rescue it.

ANALYTICS.md:71 says the global "should still be provided by GTM's Google tag". It isn't. The PR should say the swap breaks assistant events until #2307 lands, and that item needs rewording. Landing the widget change first would avoid the gap altogether.

### 3. docs.celo.org is not behind Celo's Cloudflare zone

From both 1.1.1.1 and 8.8.8.8, `docs.celo.org` is a plain CNAME to `cname.vercel-dns.com`, and the A records behind it (`76.76.21.x`, `66.33.60.x`) belong to Vercel. A proxied record returns Cloudflare IPs and hides the CNAME target; a visible CNAME to Vercel means the record is DNS-only. The `cf-ray` and `cf-cache-status` headers come from somewhere upstream of Vercel, not from the `celo.org` zone, so AI Crawl Control on that zone would see nothing.

ANALYTICS.md:82 and :84 should be unticked and the "already in place" sentence dropped. The claim at ANALYTICS.md:97 that MCP volume "is visible in Cloudflare's HTTP analytics" goes with it.

### 4. ANALYTICS.md will be a public page

ANALYTICS.md:3 says the file isn't public because it isn't in navigation. Mintlify serves every `.md` under the content root regardless: `https://docs.celo.org/AGENTS` and `/CLAUDE` are live today with full content, and `/ANALYTICS` will be too. Nothing in the file is secret, both IDs are already in page source, but the sentence is wrong. Either accept that it's public and say so, or move the content somewhere Mintlify won't serve.

### Smaller things

- The published container still holds only the Google tag and the history listener. Pushing `add_network_click` on the GTM page produced no hit, so the `AddNetworkButton.jsx` events ship dormant. That matches your comment; just confirming it's still true at the current head.
- The PR checklist still shows the ownership table as open. It's filled in as of e7fe61af.

Everything else I checked holds up. The container carries `G-0CXEKQ81V2` and sends a page view on load. `integrations.gtm.tagId` matches Mintlify's schema. The widget event names, the Redis list and its trim size match docs-ai-assistant. The plan and price claims match #2250. Google's channel-group doc lists exactly ChatGPT, Gemini, DeepSeek, Copilot and Grok for the built-in AI Assistant channel. The snippet change is clean.

GigaHierz:
All four are corrected. I re-verified the two I could check from here rather than taking them on trust, and both hold.

### 1. The History Change trigger double-counts — unticked and reversed

The runbook said the opposite of what is true and had it ticked off, which is the worst combination. It now says: **All Pages only, do not add a History Change trigger**, with the reason (GA4 enhanced measurement already sends a `page_view` on `pushState`, so the trigger fires the Google tag a second time). Your measured 1-vs-2 per navigation is quoted.

Flagging clearly: **the published container still carries that trigger.** I cannot change the GTM container from here — that is a UI action someone has to take before this item can be ticked. The file now says so rather than implying it is done.

### 2. The swap silences the assistant — stated plainly

Rewritten from a "re-verify after the swap" step into a statement of fact: `window.gtag` is a function today via `integrations.ga4` and `undefined` under the container alone, so all seven `assistant_*` events stop on merge and stay stopped until the widget pushes to `dataLayer` directly (#2307). I included your finding that a hand-rolled `gtag(){ dataLayer.push(arguments) }` shim does **not** rescue it, because that is the non-obvious part and the next person will otherwise try it. The file now recommends landing the widget change first.

### 3. Cloudflare — confirmed, and unticked

```
$ dig +short docs.celo.org @1.1.1.1
cname.vercel-dns.com.
76.76.21.93
66.33.60.194

$ curl -sI https://docs.celo.org/ | grep -i 'server\|cf-ray'
cf-cache-status: HIT
cf-ray: a4057db15a5c3851-CDG
server: Vercel
```

A visible CNAME to Vercel with Vercel addresses behind it is proof the record is DNS-only — a proxied record returns Cloudflare addresses and hides the target. The `cf-ray` header sits upstream of Vercel, on the same response that says `server: Vercel`. Runbook 3 now opens by saying none of it is in place, and everything under it is explicitly conditional on a DNS change someone has to decide on. The MCP-volume claim in Known blind spots went with it.

### 4. ANALYTICS.md will be public — confirmed and owned

```
$ curl -s -o /dev/null -w '%{http_code}' https://docs.celo.org/AGENTS   -> 200
$ curl -s -o /dev/null -w '%{http_code}' https://docs.celo.org/CLAUDE   -> 200
```

Both serve full content. The header now says the file is reachable at `/ANALYTICS`, explains that not being in `navigation` only keeps it out of the sidebar and search, and adds the instruction that follows from it: do not put anything in here you would not publish.

### The one thing I did not change

The `AddNetworkButton.jsx` events still ship dormant — the six event tags do not exist in the container. That is unchanged and still true at this head, as you said. It is harmless (a `dataLayer` push with no tag listening is a no-op) but it does mean this PR delivers no new measurement on its own.

Which raises the honest question for you as reviewer: **given the container still has the wrong trigger and none of the event tags, is there a reason to merge this now rather than after the GTM work?** The swap's only immediate effect on the live site would be to silence the assistant events. I have left it in scope because that was the call on this PR, but I would take the argument that the `docs.json` swap should be split out and land last.

### Verification

```
$ npx mintlify@4.2.920 validate                        -> build validation passed
$ npx mintlify@4.2.920 broken-links --check-redirects  -> no broken links found
$ bash scripts/check-orphans.sh                        -> No orphan pages found
```

CI here stays red until #2335 lands — `npx mintlify` currently resolves to an uninstallable release, unrelated to this diff.

Re-requesting review.

PR #2289, opened 2026-08-27
docs: rename Infra Partners to Operate, fold Specs in as a Specification section

palango:
Node operators, RPC providers, and custodians had their docs split across two tabs: "Infra Partners" held the run-a-node guides and network notices, "Specs" the protocol specification. This renames the tab to Operate, moves `infra-partners/*` to `operate/*` and `specs/*` to `operate/specification/*`, and folds the Specs tab's two sidebar groups into Operate. Redirects follow the standard set by #2279 and used in the Legacy-tab removal (#2280): wildcards cover the old paths, and every existing destination is re-pointed at its final page so nothing chains. Closes #2260.

## Changes

- All 39 pages move with `git mv` (23 infra-partners, 16 specs). The Operate tab keeps the existing groups and appends the two from Specs: Network Notices (with Archive), Node Operation, Node Management, Reference, Specification (with Deployments), Protocol Upgrades (with L2 Migration).
- Redirects: 61 destinations under `/infra-partners/` and 10 under `/specs` re-pointed to the new paths, and four entries added at the end of the array: `/infra-partners/:slug*` → `/operate/:slug*`, `/specs/:slug*` → `/operate/specification/:slug*`, plus bare `/infra-partners` → `/operate/notices/overview` and `/specs` → `/operate/specification`. The wildcards sit after the 13 pre-existing `/infra-partners/*` sources (the six notice aliases into the archive, and the integration paths), which keep matching first.
- About 140 inbound links rewritten across 47 files, plus the hardcoded FAQ anchor in `navigation.global.anchors`, the home-page card in `snippets/home.jsx`, the AGENTS.md directory list and canonical-fact table, and CODEOWNERS collapsed to a single `/operate/ @celo-org/blockchain` entry.
- Two scope items from #2260 were already done on `main`: `tooling/nodes/run-a-celo-node` was a four-line external-link stub deleted in #2237 (redirect already in place), and the #2254 "keep → Operate" set is only the operator FAQ, which was already in the tree.

## specs.celo.org

The redirect stubs (celo-org/specs#195) hard-code `docs.celo.org/specs/...` targets and forward URL fragments, so old specs.celo.org links now take one extra hop through the new `/specs/:slug*` wildcard. celo-org/specs#199 tracks re-pointing the stubs at `/operate/specification/...` after this deploys; until then the `/specs` redirects here must not be removed.

## Verification

```
$ npx mintlify broken-links
success no broken links found

$ bash scripts/check-orphans.sh
[...]

Issue #2266, opened 2026-08-21
Epic: Docs restructure — 7 tabs → Learn / Build / Operate / Contribute


GigaHierz:
### Why

The current seven-tab structure (Home, Build on Celo, Tooling, Contribute to Celo, Infra Partners, Specs, Legacy) mixes audiences, jobs and topics on one axis. Verified at `bdf40b37`: "Home" is protocol reference with no real homepage; Tooling holds 90 of 292 pages (31%) across ten unrelated groups; Legacy conflates dead L1 content with pages that still apply on L2; 105 duplicate files sit in `_deprecated/`; 25 tracked pages are unreachable from navigation (including the whole dev-environment-setup group); wallets, fee abstraction and thirdweb are each documented in three or more places; 167 redirects dead-end.

### Decisions (agreed — not re-opened in the child issues)

| Decision | Outcome |
|---|---|
| Tab shape | 7 → 4: **Learn / Build / Operate / Contribute** |
| Specs tab | Folded into Operate as a "Specification" section; redirects from old specs paths |
| Legacy tab | Removed, not archived. Still-relevant content migrated to Learn (history) or Operate (old L1 node context) before deletion |
| Tooling tab | Dissolved into Build |
| Build-tab order | Quickstart → Agents → Mini Apps → Network info → Guides → Tools → Reference |
| thirdweb | Stays as one tool page at parity with other tools; no code examples, no recommendations |
| MiniPay | One overview that routes to docs.minipay.xyz for the mini-app lifecycle |
| Code examples | Troubleshooting-first: keep edge-case examples (e.g. paying gas in USDC); drop large end-to-end examples derivable from SDK docs |
| Writing standard | `AGENTS.md` at repo root: headings, structure, writing style, move/redirect checklist |
| Highest priority now | Analytics + an AI assistant (not Mintlify Pro; research Mintlify-compatible and open-source options) |
| End-user vs tooling | Every page states who it is for; end-user project listings also live on celo.org/ecosystem |
| Orphans | Sheet of the orphaned pages only; team marks re-nav/delete; applied in one pass |
| Startup Pathway link | Not re-added (removed with Celo Camp in #2235) |

### Children, in execution order

| # | Ticket | Owner | Depends on | Status |
|---|---|---|---|---|
| #2249 | Add Google Analytics (GA4) | @viral-sangani | — | **done** — PR #2267 merged `453efdc2` |
[...]

PR #2308, opened 2026-09-02
docs(operate): add a landing page for the Operate tab

palango:
The Operate tab has had no landing page since #2289 folded the Specs tab into it, so `/operate` resolved to `operate/notices/overview` and the first screen a prospective node operator saw was a list of deprecations. This adds `operate/index.mdx` as the tab's front door and puts it first in the nav, matching every other tab, all of which already open on their own index page. The page routes rather than explains: `operate/operators/overview.mdx` keeps its role as the node-operation intro and is not touched, so node tiers, client choice and the op-geth switch dates still live in one place each.

## What's in it

`operate/index.mdx`, written to the AGENTS.md §3 Index page type: an audience sentence, a warning that op-geth is being replaced by op-reth which links out for the dates rather than restating them, then card groups for Run a node, Keep a node healthy, Reference, Protocol specification and Network notices. It closes with `## Related` pointing at the three things readers need that live outside the tab: network information, hosted RPC providers, and community RPC nodes.

`docs.json` gains a four-line insert, `{"group": "Introduction", "pages": ["operate/index"]}`, at the head of the Operate tab. Every existing group keeps its position, so Network Notices stays second because the op-geth deadline is still live. A one-page group is the existing precedent: the Home tab opens the same way.

No redirects. Nothing is moved, renamed or deleted, so AGENTS.md §2 does not apply. There is no `/operate` entry among the 737 redirects and no wildcard that would shadow it.

## Verification

```
$ mint broken-links
success no broken links found

No orphan pages found.
```

`git diff docs.json` is 4 insertions and 0 deletions, with no reformatting.

## One thing to look at

The page ends with `Ask in #celo-L2-support on Discord if you get stuck.` The equivalent line on `operate/operators/overview.mdx:50` reads "Please reach out to our team on Discord in the #celo-L2-support channel if you have any questions." I did not copy it, on the grounds that AGENTS.md says an existing page that conflicts with the standard is wrong and should not be copied as drift. That leaves the two pages wording the same pointer differently until the older one is tidied. Happy to match the old wording instead if you would rather they agree today.

Part of  #2266.


GigaHierz (changes requested):
Reviewed at head `692b8706`, rebased onto current `main` first (the branch was 2 behind; it rebases clean). Content tier — no code or money paths — so instead of a suite pass I did link resolution, nav reachability, independent verification of every claim in the body and on the page, and a rendered click-through.

One finding, one line. Everything else verified clean.

## Finding

**Low · `operate/index.mdx:89` · content sits after `## Related`, which AGENTS.md §3 says closes the page.**

The Discord line follows the `## Related` bullets. AGENTS.md: *"`## Resources` is a two-column table… `## Related` is a bullet list… Both close the page."*

I checked whether the standard is already drifting rather than assuming: of 24 pages carrying `## Related` on `main`, 22 end there. This would be the third exception.

Raising it only because the body invokes AGENTS.md to justify the *wording* of that exact sentence, so the standard is being applied to its content but not its placement.

**Fix:** move the sentence above `## Related`, or fold it into the opening paragraph beside the "fastest path" line.

## Your open question

Don't match `operate/operators/overview.mdx:50`. Yours is shorter and says the same thing, and AGENTS.md is explicit that the conflicting existing page is the one that's wrong. The right resolution is fixing the older four, not degrading the new one — the long form appears identically on `operators/overview.mdx:50`, `troubleshooting.mdx:44`, `notices/archive/l1-fusaka-upgrade.mdx:39` and `notices/archive/celo-sepolia-launch.mdx:91`. Say the word and I'll file it as a follow-up.

## A finding I withdrew

I nearly raised that the "Run a node" deck omits `/operate/operators/architecture`, which is second in the nav group this page routes. Then I checked: `operators/overview.mdx:17` links to it directly, so it is one hop from your first card. Correct curation, not a gap.

## Verified good

- **All 19 internal links** resolve to a real file *and* are reachable from nav. No anchors used, so nothing `broken-links` cannot see.
- **Every factual claim on the page**, against the pages it describes: the op-geth→op-reth switch dates exist on the deprecation notice (Sepolia June 24 2026, Mainnet July 22 2026 — correctly *not* restated here, per §7); `run-node` is op-reth-from-snapshot; `historical-proofs` is `eth_getProof` without an archive node; `configuration` is `celo-l2-node-docker-compose`.
- **Every structural claim in the body**, independently: `/operate` did resolve to `operate/notices/overview`; Operate was the only tab not opening on an index page; the Home tab is a real one-page-group precedent; 737 redirects with no `/operate` source and no wildcard shadowing it.
- **`docs.json` is exactly 4 insertions, 0 deletions**, no reformatting — diffed rather than taken on trust. Every existing group keeps its position.
- **Rendered click-through** on `mint dev`: `/operate` returns 200 and serves the new page. `broken-links` and `check-orphans` both green against current `main`, not just the base this was written on.
- The Discord channel deep-link resolves (200) and matches the pattern on five other pages.
- `sidebarTitle: "Start here"` is consistent with `build-on-celo/index`'s "Getting started".

## What it got right

The diagnosis is the valuable part — a prospective node operator's first screen being a deprecation list is a real problem, and the fix is four lines of nav plus one page. Routing rather than explaining is the right call: `operators/overview.mdx` keeps its single-source role and is untouched, so node tiers and client choice still live in one place each. Linking out for the op-geth dates instead of restating them is §7 done properly, and it is the detail most landing pages get wrong.

One round of targeted fixes, not a rethink.

## Merge-order hazards

None. `git merge-tree` against all six other open PRs (#2288, #2295, #2297, #2299, #2300, #2304) reports no conflicts. Your `docs.json` insert is in the Operate nav block; the others are in Contribute nav, redirects, footer and integrations. Merge in any order.

palango:
Fixed in `7e7ed84e`. You were right, and the cause was upstream of the page: the task brief I wrote specified the Discord line after `## Related`, so the placement was baked in before a word of the page existed.

I took your second option and folded the sentence into the opening paragraph rather than leaving it floating above `## Related`. On a landing page "who this is for, the fastest path, where to ask" reads as one thought, and it avoids an orphan sentence after the last card group. The page now ends on the Related bullets. I also checked the repo the way you did rather than assuming: no page pairs a help line with `## Related` at all, so there was no competing convention pulling the other way.

`broken-links` and `check-orphans` are green on the new head. Pushed as a plain commit rather than a rebase, so your review is intact.

On the Discord wording, agreed, and thanks for turning up the other three. I only knew about `operators/overview.mdx:50`. Filing the tidy-up as a follow-up sounds right if you are up for it, since fixing the older four is the resolution that actually holds the standard.

The withdrawn architecture finding was the right call, and for the reason you landed on: `operators/overview.mdx` already routes there, so a second path from this page would have been noise.

GigaHierz (approved):
Approving. Re-checked at head `16513d72` (the merge from `main`), not at the head I reviewed yesterday.

The one finding is closed. `7e7ed84e` moves the support sentence into the opening paragraph, so the page now ends on the `## Related` bullets and matches the 22 of 24 pages that do the same. Folding it into the "who this is for, fastest path, where to ask" sentence reads better than the floating line would have.

Verified on this head:

- Diff is still exactly `docs.json` +4 and `operate/index.mdx` +87, nothing else, no reformatting.
- `mint broken-links` and `bash scripts/check-orphans.sh` both green locally on `16513d72`.
- `mint dev` serves `/operate` with a 200 and the rendered page carries the Start here title, the support sentence in the intro, and a single `## Related` at the end.
- Branch is 0 behind `main`, CI green, no conflicts with the other open PRs.

The Discord wording tidy-up on the older four pages is filed as #2309.

PR #2316, opened 2026-09-08
docs(nodes): add BoltRPC as Celo RPC provider

KubilayJackson:
## Summary

- Add BoltRPC to the As a Service RPC provider list on `/tooling/nodes/overview`
- Celo Mainnet only (full archive, HTTPS, WebSocket, debug methods); operated by Matrixed.Link

## Test plan

- [ ] Name link opens https://boltrpc.io
- [ ] Operator link opens https://matrixed.link
- [ ] Card opens https://boltrpc.io/networks/celo
- [ ] Section renders after NOWNodes under As a Service

### DIFF
--- a/tooling/nodes/overview.mdx
+++ b/tooling/nodes/overview.mdx
@@ -220,3 +220,15 @@ https://getblock.io/nodes/celo/
 <Card href="https://nownodes.io/nodes/celo-celo" arrow title="NOWNodes">
 https://nownodes.io/nodes/celo-celo
 </Card>
+
+### BoltRPC
+
+[BoltRPC](https://boltrpc.io) is a multi-chain RPC provider operated by [Matrixed.Link](https://matrixed.link). Celo Mainnet is served as a full archive node with HTTPS, WebSocket, and debug methods. Operations are ISO/IEC 27001:2022 certified.
+
+#### Supported Networks
+
+- Celo Mainnet
+
+<Card href="https://boltrpc.io/networks/celo" arrow title="BoltRPC">
+https://boltrpc.io/networks/celo
+</Card>


GigaHierz (approved):

GigaHierz (changes requested):
Thanks for this, and apologies for the slow turnaround.

I did the provider check and **Matrixed.Link clears the bar** — the Chainlink ecosystem directory lists you as a node operator from April 2022, the Injective validator announcement checks out, and the domain dates to 2021. The BoltRPC Celo page is substantive rather than a templated stub, and the chain facts on it (42220, the L1→L2 migration, fee abstraction) are accurate. So this belongs on the list.

Three things before it can merge.

### 1. Please cut the ISO/IEC 27001:2022 sentence

I could not corroborate it. `matrixed.link/trust` states the certificate is effective 2026-02-03 with scope "available under NDA", but there is **no certificate number, no issuing registrar, and no accreditation-registry entry** I can find — every search result asserting the certification traces back to your own site. There is also an internal inconsistency worth fixing on your end regardless: the site-wide header badge reads "ISO 27001:2022 Operator since 2021" while the FAQ dates the certificate to February 2026.

To be clear about why this is a hard no rather than a nitpick: **no other entry on that page makes a compliance claim.** If we publish one we cannot verify, we have to police every one that follows. Happy to reinstate the sentence if you can give me a certificate number and the issuing body.

### 2. Please rewrite the archive sentence

The diff says Celo Mainnet "is served as a full archive node with HTTPS, WebSocket, and debug methods". Your own Celo page walks that back further down:

> Archive-only methods are available on request for qualifying plans.
> ARCHIVE ON REQUEST (3): `eth_getBalance`, `eth_call`, `debug_traceTransaction`

and the pricing page gates it from Business ($699/mo) upward. A Builder-tier reader would not get what the sentence promises. Something like:

> BoltRPC is a multi-chain RPC provider operated by Matrixed.Link. Celo Mainnet endpoints support HTTPS, WebSocket, and debug methods; archive access is available on request for qualifying plans. Paid plans only, with a 14-day trial.

I also could not test any of the capability claims — `endpoints.boltrpc.io/rpc/celo` returns 401 and there is no keyless endpoint, so `eth_chainId`, an early-block `eth_getBalance`, and `debug_traceBlockByNumber` all went unverified. If you can share a demo endpoint or a temporary key I'll run those and we can state the archive support with more confidence rather than less.

### 3. Please rebase — the branch conflicts

Three provider entries (node101, PublicNode, and one earlier) landed at the end of the same section since you branched, so this appends to a spot that has moved. The resolution is trivial: put the BoltRPC block after PublicNode at the end of "As a Service".

While rebasing, one style point: the page has drifted between `#### Supported Networks` and `#### Supported networks`. Our style guide asks for sentence case on new headings, so `#### Supported networks`.

### One thing worth raising, not blocking

There's no status page — `status.boltrpc.io` doesn't resolve and `/status` 404s. Every other as-a-service provider on that page has one. Not a blocker for the listing, but it is the first thing a developer evaluating an RPC provider looks for.

Ping me once it's rebased and I'll re-review straight away.

PR #2317, opened 2026-09-08
docs(exchanges): add EIDEX to the CELO exchange list

slaenov:
## The hole, and the fix

The list covers centralized venues, DEXs and Celo-native swaps, but there is nothing that shows what different providers actually quote before you pick one.

Adds EIDEX under Decentralized Exchanges, right after the existing "Other Exchanges" line. You put in an amount, it shows live CELO quotes from the providers side by side, sorted by what you get out. No account, no signup - the swap itself happens on the provider's own site.

## What this does NOT do / residual risk

One line of docs. No code, no config, nothing in the build.

Right now the linked page resolves CELO to USDT through three providers (ChangeNOW, StealthEX, Exolix). So it is not "every exchange", only the ones that quote CELO.

## Judgement calls

Put it under Decentralized Exchanges instead of adding a heading, since "Other Exchanges" already sits there and that one is a comparison link too, not a DEX. Move it wherever you prefer.

## Issues

None.

## Stacking / conflicts

Off main, no other PRs open from me. One file.

## Verification evidence

Opened the page and checked it returns quotes at 10, 100 and 1000 CELO.
Loads without an account.
No tests - docs only.

## Remaining ops steps

- [ ] none

## Checklist

- [x] Title is the commit message I want on main
- [x] Judgement calls flagged above
- [x] No secrets in the diff

Questions for the maintainer: none.

### DIFF
--- a/home/exchanges.mdx
+++ b/home/exchanges.mdx
@@ -26,6 +26,7 @@ Be sure you understand and review the risks when swapping assets.
 - [Uniswap](https://app.uniswap.org/)
 - [Velodrome](https://velodrome.finance/)
 - [Other Exchanges](https://coinmarketcap.com/currencies/celo/)
+- [EIDEX](https://eidex.io/screener/celo-celo-to-usdt-erc20) - Compares live CELO swap quotes across exchange providers
 
 #### Celo Specific Exchanges
 


GigaHierz:
Thanks for putting this together, and for the clear writeup of what the linked page does.

We're going to pass on this one, and I want to give you the actual reasoning rather than a one-liner.

**The bar for this page is deliberately narrow.** It lists a small number of well-known venues plus one neutral price reference, because it is a page that sends people somewhere to move real money. Every current entry is a named, established operator.

**The linked page is monetised on the providers it ranks.** The screener URL in this PR ships a `referrals` map in its own page payload, including two of the three providers you named:

```
"stealthex":"/?ref=KAriUnX44n"
"exolix":"?ref=BCDE0B79B01649638F049371D111603D"
```

plus eleven more. To be fair about it: the site *does* state elsewhere that providers share a referral reward and that routes are ranked by best output — that is a creditable disclosure. But it does not render on the screener page itself, so a reader arriving from these docs sees ranked quotes and outbound buttons with no indication that the ranking party is paid by the ranked parties.

That is also the answer to the CoinMarketCap comparison in your PR body, which was a fair point to raise: CoinMarketCap is not paid by the venues it lists and does not append affiliate codes to outbound links. That difference is the whole decision here.

**The categorisation doesn't hold either.** ChangeNOW, StealthEX and Exolix are custodial instant-swap services, not DEXs, and no swap settles on-chain on the linked page — the page says so itself ("Quotes are indicative, not executable").

None of this is a judgement about the service, and it isn't a permanent no to you as a contributor. If there is a genuine Celo integration you'd like documented elsewhere in these docs, open something and we'll look at it on its merits.

One thing this did surface on our side: the existing Binance link on the same page carries a `?ref=` parameter. That is inconsistent of us, and we'll clean it up separately — it doesn't change the outcome here, but you'd have been right to point at it.

